PR #4, opened 2026-03-30
Metaldata Proposal

maxmoehl:
With this proposal we would like to introduce a metadata service to serve instance specific metadata to a server. The goal is to provide a generic mechanism that works similar to how the cloud service providers are doing metadata services to ultimately decouple us from the tight integration with the ignition ecosystem and allow other configuration tools to use the generic service with ignition being one of them.

### DIFF
--- a/ieps/18-metaldata.md
+++ b/ieps/18-metaldata.md
@@ -1,0 +1,218 @@
+---
+title: Metaldata Service
+
+ep-number: 18
+
+creation-date: 2026-03-30
+
+status: implementable
+
+authors:
+
+- "@maxmoehl"
+
+reviewers:
+
+- "@afritzler"
+- "@hardikdr"
+
+---
+
+# IEP-18: Metaldata Service
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Alternatives](#alternatives)
+
+## Summary
+
+Introduce a new component to the IronCore metal-automation which serves metadata
+information to servers.
+
+## Motivation
+
+In order to configure a server properly the script configuring it needs to be
+aware of which server it is configuring to choose the correct host name, FRR
+configuration and other per-server properties.
+
+One way to achieve this is to create per-server ignition (or cloud-init) scripts
+which are already tailored to the host they are supposed to run on. However,
+sometimes these scripts can not be fully customized as they are re-used for a
+set of servers. The workaround so far was to create an inventory file which
+contains the per instance data using the system UUID as a key. A startup script
+would then look at that inventory file, pick up the right metadata and fill in
+the gaps. This is a cumbersome process as it requires collecting all system
+UUIDs upfront, the scripts are not reusable across environments, and every
+server has access to all the metadata.
+
+To alleviate this issue we propose to introduce a meta(l)data service.
+
+### Goals
+
+* Serve metadata information: hostname, zone, IP prefix.
+* Serve user-data in a key-value format.
+* Version the endpoints to support breaking changes in the future.
+
+### Non-Goals
+
+`¯\_(ツ)_/¯`
+
+## Proposal
+
+Implement a meta(l)data service.
+
+The metadata should be presented as JSON by default, support for additional
+formats may be added later on. The client can request a specific format using
+the `Accept` HTTP header. The response should look like this:
+
+```json
+{
+    "server-name": "cn-wdf4a-1-system-0",
+    "hostname": "cn-wdf4a-1",
+    "zone": "wdf4a",
+    "ip-prefix": "2001:db8::/32",
+    "user-data": {
+        "foo": "bar"
+    }
+}
+```
+
+### Data Source
+
+The top-level keys are populated based on the annotations and labels of the
+server in the following format:
+
+```yaml
+metadata.metal.ironcore.dev/$key: $value
+```
+
+This allows the operator to specify arbitrary key-value pairs in the returned
+metadata JSON. When an annotation and label with the same key are specified, the
+label takes precedence.
+
+In addition to those key-value pairs, the name of the server objects is also
+included in the field `server-name`. Specifying an annotation or label with
+that key overwrites it with the given value.
+
+For data that is common among all servers (e.g. the region or zone) the operator
+can define static key-value pairs which are set on every response. Those can
+also be overridden by annotations / labels and the key `server-name` cannot be
+specified.
+
+Beyond the server annotations / labels a field for user data will be
+added. This allows us to progress towards feature-parity with metadata services
+offered by similar platforms. The [list of supported
+platforms](https://coreos.github.io/ignition/supported-platforms/) for ignition
+shows that it is a common pattern to use such a field to deliver the ignition
+script. This has the added benefit that we are no longer hard-wired to the
+ignition ecosystem as this approach could be adapted to support other systems
+like cloud-init.
+
+User-data can be supplied via the `UserDataRef` of a `ServerClaim`. The
+metadata server resolves the claim bound to a server to get the secret
+referenced by the claim containing the user-data.
+
+### Metaldata API
+
+To retrieve data from a local metadata server, clients rely on DNS to resolve to
+the correct local instance:
+
+```
+metaldata.ironcore.dev
+```
+
+The metadata service uses the client IP to identify the server which called it.
+The metadata is extracted from the resources which are populated by the
+metal-operator, specifically the `Server` custom resource.
+
+Depending on the path different information is returned:
+
+* `/v1/`: the full metadata JSON, similar to the example above.
+* `/v1/{field}`: the content of the named field as plain text, e.g. `cn-wdf4a-1` if
+  the user called `/hostname` in our example.
+* `/v1/user-data`: the user-data as JSON.
+* `/v1/user-data/{key}`: the value of the user-data key as plain text.
+
+If a key contains special characters not allowed in URLs, standard URL encoding
+must be applied to retrieve them.
+
+This allows us to stay compatible with the `ignition.config.url` kernel cmdline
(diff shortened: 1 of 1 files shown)


ieps/13-metaldata.md
```diff
@@ -0,0 +1,216 @@
...
+	Image string `json:"image"`
+
+	// UserDataRef references a Secret object which contains the user-data which
+	// will be served via the metaldata service to the server.
+	// +optional
+	UserDataRef *v1.LocalObjectReference `json:"userDataRef,omitempty"`
+}
+```
+
+No changes to metal-operator itself are required at this time as this
```
afritzler: Maybe we should rephrase this to: Apart from the API change no behavioral changes in the controller logic is necessary at this this point in time.

afritzler (changes requested):

ieps/18-metaldata.md
```diff
@@ -0,0 +1,216 @@
...
+    "ip-prefix": "2001:db8::/32",
+    "user-data": {
+        "foo": "bar"
+    }
+}
+```
+
+### Data Source
+
+The top-level keys are populated based on the annotations and labels of the
```
hardikdr, line 87: Is there any specific reason for including both labels and annotations? Could we use only either of them, to avoid extra logic to prevent duplicated entries?

ieps/13-metaldata.md
```diff
@@ -0,0 +1,216 @@
...
+
+This allows the operator to specify arbitrary key-value pairs in the returned
+metadata JSON. When an annotation and label with the same key are specified, the
+label takes precedence.
+
+In addition to those key-value pairs, the name of the server objects is also
+included in the field `server-name`. Specifying an annotation or label with
+that key overwrites it with the given value.
+
+For data that is common among all servers the operator can define static
```
hardikdr: Could we simplify this statement, with an example, maybe I couldn't understand it.

ieps/18-metaldata.md
```diff
@@ -0,0 +1,216 @@
...
+ignition ecosystem as this approach could be adapted to support other systems
+like cloud-init.
+
+User-data can be supplied via the `UserDataRef` of a `ServerClaim`. The
+metadata server resolves the claim bound to a server to get the secret
+referenced by the claim containing the user-data.
+
+### Metaldata API
+
+To retrieve data from a local metadata server, clients rely on DNS to resolve to
```
hardikdr, line 122: It should also support hard-coded reserved IP to avoid dependency on functional DNS imo.

ieps/18-metaldata.md
```diff
@@ -0,0 +1,216 @@
...
+service does not implement HTTPS.
+
+We explicitly do not implement the approach of IDMSv2 from AWS which uses a
+token based flow. This is done to allow for simple consumption in environments
+where only very limited tooling is available (like the initrd phase). A future
+version may support such a mechanism as an opt-in feature.
+
+### Metal-API Changes
+
+As outlined earlier we will have to add support for a user-data secret to the
```
hardikdr, line 170: We might want to quickly think, if there are any potential overlaps with the ongoing efforts related to `ServerMetadata.` 
Ref: https://github.com/ironcore-dev/metal-operator/pull/762/changes

ieps/18-metaldata.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Goals
+
+* Serve metadata information: hostname, zone, IP prefix.
+* Serve user-data in a key-value format.
+* Version the endpoints to support breaking changes in the future.
+
+### Non-Goals
+
+`¯\_(ツ)_/¯`
```
hardikdr, line 63: 👀

hardikdr (commented):
Thanks @maxmoehl 
I had a quick first pass and have dropped a few comments inline.

ieps/18-metaldata.md
maxmoehl: You will probably have labels for being able to include things like topology information, but on the other hand you also have data that does not fit the restricted format of labels und thus have to go into an annotation (e.g. IPv6 addresses).

ieps/13-metaldata.md
maxmoehl: An example would be the region or zone, which is the same of all servers inside a cluster.

ieps/18-metaldata.md
maxmoehl: Magic IPs are rather something for virtualized environments in which you can have SDN mangle the request. But in our bare-metal hardware setup we have to work with a real network which makes DNS the natural choice. In the end, you can also specify the IP address, it's up to the client how to discover the address of the server.

ieps/18-metaldata.md
maxmoehl: There are some touch-points but these two efforts are mostly separate. Metaldata consumes that information but supplying user-data is a separate concern which should be a property of the claim.

ieps/13-metaldata.md
maxmoehl: Addressed.

ieps/13-metaldata.md
maxmoehl: Added a note for the example.

afritzler (approved):

ieps/13-metaldata.md
```diff
@@ -0,0 +1,218 @@
...
+- "@maxmoehl"
+
+reviewers:
+
+- "@main-reviewer-1"
+- "@main-reviewer-2"
+
+---
+
+# IEP-1: Metaldata Service
```
hardikdr: Nit: IEP-13

ieps/13-metaldata.md
```diff
@@ -0,0 +1,218 @@
...
+```
+Metadata-Flavor: IronCore Metal
+```
+
+The metaldata service assumes that the network over which the request is made is
+secure. IP addresses must be trustworthy and only operators have direct access
+to the network devices which handle the packets to prevent eavesdropping as the
+service does not implement HTTPS.
+
+We explicitly do not implement the approach of IDMSv2 from AWS which uses a
```
hardikdr: Nit: IMDSv2

ieps/18-metaldata.md
```diff
@@ -0,0 +1,218 @@
...
+	IgnitionSecretRef *v1.LocalObjectReference `json:"ignitionSecretRef,omitempty"`
+
+	// Image specifies the boot image to be used for the server.
+	// +required
+	Image string `json:"image"`
+
+	// UserDataRef references a Secret object which contains the user-data which
+	// will be served via the metaldata service to the server.
+	// +optional
+	UserDataRef *v1.LocalObjectReference `json:"userDataRef,omitempty"`
```
hardikdr, line 204: Having 2 APIs, one based on Ignition, and one based on Userdata might be a confusion UX for the MetalAPI users, specifically if we dont have a buy-in from the existing adopters. I would suggest a review from a wider-audience.

hardikdr (changes requested):
Quick 2 cents, optional suggestion:
- We are essentially trying to implement the IMDSv1 model from AWS. I would recommend we should try to think of a more stable IMDSv2-like design here from the start. This can lead to a similar mess AWS has at the moment of migrating tenants from IMDSv1 to v2. 
- Nevertheless, the ClientIP-based approach is a good starting point, and build on top of it the token-based approach later. For now, it would be nice to document a possible future approach in the IEP. 
- Given that we are proposing this as an enhancement over an already working system, this might be more useful if we keep the security in mind from the start.

ieps/13-metaldata.md
maxmoehl: Can you please elaborate why you disagree with the reasoning provided in the proposal?

ieps/18-metaldata.md
maxmoehl: That is exactly the purpose of this public proposal. The two APIs are needed to support a smooth transition from the old to the new API.

ieps/13-metaldata.md
hardikdr: I meant here the typo, `s/IDMSv2/IMDSv2/`

ieps/18-metaldata.md
maxmoehl: To be clear: the idea is to support a migration by temporarily offering both APIs but eventually dropping the ignition-only path.

defo89:
I would like to raise a question whether a single `UserDataRef` field should cover two use cases. From my understanding of the cloud providers including AWS, the usual approach is:

- Metadata - key-value pairs that provides properties about the instance. In `metal-operator` case this serves instance specific metadata to a server (metaldata service).

- Userdata - custom configuration provided by user in a well-defined format (ignition, cloud-init). Metadata can be leveraged here (example: fetch server name in the init script). 

This approach may have the following benefits:
- `UserDataRef` can act as a drop-in replacement of ignition-specific `IgnitionSecretRef`. Possibly if the `type` is specified in the secret, it can be cloud-init or ignition format configuration. But this still allows to provide the user configuration as before.
- As specified in the example `Userdata` can be part of `Metadata` key-value pair and allow for even more flexible deployment options.

hardikdr (approved):

maxmoehl:
> * Userdata - custom configuration provided by user in a well-defined format (ignition, cloud-init). Metadata can be leveraged here (example: fetch server name in the init script).

I'd argue that from AWS perspective the user-data is not well-defined, it's just an opaque block of data that the user can specify. In this case it was easier to use the existing secrets type to re-use existing types instead of defining new ones while also allowing the user to specify multiple distinct values.

> * `UserDataRef` can act as a drop-in replacement of ignition-specific `IgnitionSecretRef`. Possibly if the `type` is specified in the secret, it can be cloud-init or ignition format configuration. But this still allows to provide the user configuration as before.

You can still achieve this behavior by providing a single `ignition` key in the user-data secret and using `/v1/user-data/ignition` to retrieve it. The result should be almost indistinguishable from how it works on AWS. This should also work as a drop-in replacement for the existing ignition flow with the important caveat that we no longer compile the ignition. But that is somewhat intentional: the goal is to get rid of any specifics for any particular system and become more agnostic.

PR #5, opened 2026-04-09
Proposal for Device Metadata in Custom Resources

peanball:
Some physical properties cannot be auto-detected for resources managed
by IronCore. This proposal provides data center relevant information that
is useful to operators as filterable labels on Kubernetes resources.

### DIFF
--- a/ieps/14-device-metadata.md
+++ b/ieps/14-device-metadata.md
@@ -45,9 +45,9 @@ that cannot be auto-determined from the resources themselves. Such information
 will need to be provided externally.
 
 Operators need to filter and select resources by physical properties — for
-example, listing all servers in a specific datacenter, rack, or of a certain
+example, listing all servers in a specific location, rack, or of a certain
 height. Visualization tools need this information to render datacenter layouts.
-The [metaldata service](proposal-0002-metaldata-service.md) can expose these
+The [metaldata service](13-metaldata-service.md) can expose these
 labels to servers at runtime.
 
 ### Goals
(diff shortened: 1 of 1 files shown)


ieps/14-device-metadata.md
```diff
@@ -0,0 +1,215 @@
...
+expose them to servers.
+
+### Labels
+
+Labels are used for metadata that operators need to filter or select on.
+
+| Label | Description | Example |
+|---|---|---|
+| `metadata.metal.ironcore.dev/site` | Site or campus identifier | `WDF4` |
+| `metadata.metal.ironcore.dev/datacenter` | Datacenter within a site | `RZ78` |
```
maxmoehl: Somehow `datacenter` doesn't quite sit right with me, I think the term is too overloaded to mean anything specific.

Since, for the most part, `site` and `datacenter` are going to be static within a single installation of metal-operator, should we just combine them? Then we could use whatever term we like as it just means "where this building / room / complex is" and the user can define the exact semantics.

ieps/14-device-metadata.md
```diff
@@ -0,0 +1,215 @@
...
+| `metadata.metal.ironcore.dev/rack` | Opaque rack identifier (may encode row, position, or other site-specific schemes) | `1-12` |
+| `metadata.metal.ironcore.dev/rack-position` | Position within the rack (bottom-up U number) | `24` |
```
maxmoehl:
```suggestion
| `metadata.metal.ironcore.dev/shelf` | Position within the rack (bottom-up U number) | `24` |
```

I think shelf is a common term to refer to the height at which a server is installed within a rack.

maxmoehl (commented):
Well written and thought out, as always.

Looking through the examples I noticed that we report the model of the server. Technically the height could be derived from that if we have a mapping somewhere, but I don't think I want to introduce that now. This is something the operator has to deal with when filling in the information, at least for now.

peanball:
> Technically the height could be derived from that if we have a mapping somewhere, but I don't think I want to introduce that now.

Yes, you're right that there would be some mapping from device type to height.

I see the following options:
1. Define a "device type" resource that is referenced based on vendor/model. The issue sometimes is with vendors being bad at naming and calling different revisions with different specs the same name. So you'd need to go arbitrarily deep in the mapping (HW revision, subtype, whatever you may need to uniquely identify this model).

  I think this would also make it more difficult to filter via those labels.

2. We shift the mapping to the operator and their tooling, which can then use their inventory system to put explicit data for the actual things for this particular device in its metadata.

3. A hybrid. We keep the map of devices to their metadata as "templates" as new CRD and use them during reconciliation. But this feels bloated in terms of approach and in terms of CRs floating around.

My preference is on 2.

ieps/14-device-metadata.md
peanball: I think this is where we might be back to `location` :)

The location is whatever precision information you need, i.e. site, building, wing, floor, room, whatever, where your cluster is hosted. We assume that the whole cluster is hosted in one `location`.

The `rack` is then the subdivision within that location that identifies a particular rack.

Scenario:
You have multiple rooms on one floor for your cluster.
* `location` is country, city, building identifier, wing?, floor (as needed)
* `rack` is e.g. room 341, rack row 17, rack 5

So the idea would be:
* `location` is the static "common" bit of information. This could even be an annotation as we are not going to filter for that.
* `rack` is the unique identifier of a particular rack within `location`.

maxmoehl: Agreed, I'd say make both a label for consistency.

peanball: Done.

maxmoehl (approved):
Let's present this alongside the metaldata proposal in the next metal community call.

afritzler (approved):

ieps/14-device-metadata.md
```diff
@@ -0,0 +1,208 @@
...
+
+The `rack` label is intentionally an opaque string. Different environments may
+use different conventions to uniquely identify a rack (e.g. row and position,
+floor and aisle, or a flat identifier). A recommended format is not prescribed
+— the value should be whatever uniquely identifies the rack within the
+location.
+
+> [!NOTE]
+> 1. Since each management cluster corresponds to a single zone / availability
+>    zone, an AZ label is not needed — it is implicit from the cluster context.
```
defo89: This may depend on the environment. A cluster that hosts `metal-api` CRs might stretch across multiple zones. Hence it might be helpful to have `zone/region` labels.

Since `cloud-controller-manager` [needs to apply this labels](https://github.com/ironcore-dev/cloud-provider-metal/blob/main/pkg/cloudprovider/metal/instances_v2.go#L145-L151) on the nodes as per [contract](https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone), maybe this information could be unified somehow.

peanball: Agreed. I will adjust the text and examples accordingly.

I will add `zone`, and `region` as "overarching" container that may contain multiple zones.

peanball: @defo89, could you have a look whether the latest state with [f25512b](https://github.com/ironcore-dev/enhancements/pull/5/commits/f25512b61012f0a29599dd7d26817776daa9cfea) is what you had in mind / works for you?

defo89: @peanball Thanks! This does indeed work for our use case.

defo89 (approved):

PR #6, opened 2026-04-13
Pool Health Proposal

adracus:
This proposal defines how to determine a pool's health and availability and which parties are involved in computing and keeping it up-to-date.

### DIFF
--- a/ieps/15-pool-health.md
+++ b/ieps/15-pool-health.md
@@ -1,0 +1,160 @@
+---
+title: Pool Health
+
+iep-number: 15
+
+creation-date: 2026-05-13
+
+status: implementable
+
+authors:
+
+- "@adracus"
+
+reviewers:
+
+- "@afritzler"
+- "@balpert89"
+
+---
+
+# IEP-15: Pool Health
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Alternatives](#alternatives)
+
+## Summary
+
+To determine the availability and health of each pool (`compute`, `storage`),
+we propose introducing heartbeats and status conditions of pools.
+
+Similar to Kubernetes' node heartbeats & status conditions, there are 2
+parties involved in this:
+
+1. The pool itself, regularly reporting its status by patching the `*Pool`
+   resource's status and acquiring / keeping a lease object within the
+   `ironcore-<type>-lease` namespace (with `type` being either `machinepool`,
+   `volumepool` or `bucketpool`).
+   For references on leases, see the [leases Kubernetes documentation](https://kubernetes.io/docs/concepts/architecture/leases/).
+2. A pool controller within the IronCore control plane that updates
+   a pool object's status conditions depending on the heartbeats of the pool.
+
+## Motivation
+
+Currently, there are no means to determine whether a pool is healthy or ready
+other than trying to interact with the pool. Additionally, features like
+pool rolling and / or eviction need proper health information of pools to
+function properly.
+
+### Goals
+
+* Require a well-defined lease per pool to be acquired / kept.
+* Expose health information on a pool's `.status.condtions`.
+* Extend / create pool reconcilers that monitor the pools and update
+  pool status if pools fail to do their heartbeat.
+
+### Non-Goals
+
+* Handle eviction - Eviction is a subsequent process that depends on proper
+  health information. It also needs to be discussed in which form machines
+  / volumes can / should be evicted.
+* Handle pool rolling - needs to be done subsequently as well.
+
+## Proposal
+
+For each IronCore installation, create the following additional namespaces:
+* `ironcore-machinepool-lease` for `MachinePool` leases.
+* `ironcore-volumepool-lease` for `VolumePool` leases.
+* `ironcore-bucketpool-lease` for `BucketPool` leases.
+
+Inside these namespaces, for each `<Type>Pool`, a pool controller
+must acquire and keep a `coordination.k8s.io/Lease`:
+For example, if there is a `MachinePool`
+`my-machinepool`, the controller of this `MachinePool` must acquire and
+keep a lease called `ironcore-machinepool-lease/my-machinepool`. The
+recommended default for setting on the lease is 40 seconds.
+
+```yaml
+apiVersion: coordination.k8s.io/v1
+kind: Lease
+metadata:
+  name: my-machinepool
+  namespace: ironcore-machinepool-lease
+spec:
+  holderIdentity: machinepoollet-07a5ea9b9b072c4a5f3d1c3702_0c8914f7-0f35-440e-8676-7844977d3a05
+  leaseDurationSeconds: 3600
+  renewTime: "2026-05-01T13:18:48.065888Z"
+```
+
+Additionally, each pool controller must regularly update a condition
+called `Ready` in a `Pool`'s `.status.conditions`. By default, an
+interval of 10 seconds is recommended (taken from Kubernetes' kubelet defaults).
+The pool controller must set this condition to `True` if the pool is ready
+for workload, otherwise it must explicitly set it to `False`.
+
+Healthy example:
+
+```yaml
+apiVersion: compute.ironcore.dev/v1alpha1
+kind: MachinePool
+metadata:
+  name: machinepool-sample
+spec:
+  providerID: ironcore://shared
+status:
+  conditions:
+  - type: Ready
+    status: True
+    reason: HeartbeatReceived
+    message: A heartbeat for the machine pool has been received within the allotted time.
+    observedGeneration: 4
+    lastTransitionTime: "2026-05-01T13:18:48.065888Z"
+```
+
+Unhealthy example:
+
+```yaml
+apiVersion: compute.ironcore.dev/v1alpha1
+kind: MachinePool
+metadata:
+  name: machinepool-sample
+spec:
+  providerID: ironcore://shared
+status:
+  conditions:
+  - type: Ready
+    status: Unknown
+    reason: NoHeartbeatReceived
+    message: No heartbeat for the machine pool has been received within the past time.
+    observedGeneration: 4
+    lastTransitionTime: "2026-05-01T13:18:48.065888Z"
+```
+
+For each pool type (`MachinePool`, `VolumePool`, `BucketPool`), a controller is
+created / reconciler amended to the `ironcore-dev` controllers (not the `*poollet*`
+that monitors the pools of that type.
+
+If a pool is unresponsive for a specific duration, its `Ready` condition
+is marked as `Unknown`. A recommended default for the
+`<type>pool-monitor-grace-period` is 50 seconds. A pool is deemed
+unresponsive if there was no progress on neither the pool's status
+nor the pool's lease.
+
+> [!IMPORTANT] The pool controller should not inspect timestamps
+> of the conditions / leases. This is prone to failure due to clock
+> skew. Instead, pool controllers should check if there were updates
+> since the last inspection and only if there were no updates for a
+> specific duration mark the pool as non-ready.
+
+`pool.status.state` will be deprecated in favor of the more fine-grained
+`Ready` condition (and potentially other future conditions).
+
+## Alternatives
+
+* Keep it as-is and work via trial-and-error.
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,12 @@
+.idea
+*.swp
+*.swo
+*~
+.vscode/
+testbin/
+dev/
+charts/
+apiserver.local.config/
+default.etcd/
+.DS_Store
+


afritzler:
Should we explicitly also list the `BucketPools` in this proposal?

adracus:
> Should we explicitly also list the `BucketPools` in this proposal?

Good catch, forgot about `BucketPool` - amended the proposal.

ieps/13-pool-health.md
```diff
@@ -0,0 +1,108 @@
...
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Alternatives](#alternatives)
+
+## Summary
+
+To determine the availability and health of each pool (`compute`, `storage),
```
friegger:
```suggestion
To determine the availability and health of each pool (`compute`, `storage`),
```

ieps/13-pool-health.md
```diff
@@ -0,0 +1,108 @@
...
+### Non-Goals
+
+* Handle eviction - Eviction is a subsequent process that depends on proper
+  health information. It also needs to be discussed in which form machines
+  / volumes can / should be evicted.
+* Handle pool rolling - needs to be done subsequently as well.
+
+## Proposal
+
+For each IronCore installation, create 2 additional namespaces:
```
friegger:
```suggestion
For each IronCore installation, create additional namespaces:
```

ieps/15-pool-health.md
```diff
@@ -0,0 +1,108 @@
...
+* `ironcore-machinepool-lease` for `MachinePool` leases.
+* `ironcore-volumepool-lease` for `VolumePool` leases.
+* `ironcore-bucketpool-lease` for `BucketPool` leases.
+
+Inside these namespaces, for each `<Type>Pool`, a pool controller
+must acquire and keep a `coordination.k8s.io/Lease`:
+For example, if there is a `MachinePool`
+`my-machinepool`, the controller of this `MachinePool` must acquire and
+keep a lease called `ironcore-machinepool-lease/my-machinepool`. The
+recommended default for setting on the lease is 40 seconds.
```
friegger, line 81: It would be nice to have a short example here, like:
```
Name:         my-machinepool
Namespace:    ironcore-machinepool-lease
API Version:  coordination.k8s.io/v1
Kind:         Lease
Metadata:
  Creation Timestamp:  <timestamp>
  Owner References:
    API Version:     v1
    Kind:            MachinePool
    Name:            my-machinepool
    UID:             <uid>
  Resource Version:  28894925
  UID:               <uid>
Spec:
  Holder Identity:         my-machinepool
  Lease Duration Seconds:  40
  Renew Time:              <timestamp>
```

ieps/15-pool-health.md
```diff
@@ -0,0 +1,108 @@
...
+
+Additionally, each pool controller must regularly update a condition
+called `Ready` in a `Pool`'s `.status.conditions`. By default, an
+interval of 10 seconds is recommended. The pool controller must set
+this condition to `True` if the pool is ready for workload, otherwise
+it must explicitly set it to `False`.
+
```
friegger, line 138: Here also a short excerpt from the status section with an example would be nice.

ieps/15-pool-health.md
```diff
@@ -0,0 +1,108 @@
...
+For each pool type (`MachinePool`, `VolumePool`, `BucketPool`), a controller is
+created / reconciler amended that monitors the pools of that type.
+If a pool is unresponsive for a specific duration, its `Ready` condition
+is marked as `Unknown`. A recommended default for the
+`<type>pool-monitor-grace-period` is 50 seconds. A pool is deemed
+unresponsive if there was no progress on neither the pool's status
```
friegger, line 146: This could be expressed clearer, unhealthy means lease not renewed and condition `Ready` not true.

ieps/15-pool-health.md
```diff
@@ -0,0 +1,108 @@
...
+nor the pool's lease.
+
```
friegger, line 148: From the above, what are the values the `Ready` condition can assume? `True`, `False` and `Unknown`?

ieps/13-pool-health.md
```diff
@@ -0,0 +1,108 @@
...
+must acquire and keep a `coordination.k8s.io/Lease`:
+For example, if there is a `MachinePool`
+`my-machinepool`, the controller of this `MachinePool` must acquire and
+keep a lease called `ironcore-machinepool-lease/my-machinepool`. The
+recommended default for setting on the lease is 40 seconds.
+
+Additionally, each pool controller must regularly update a condition
+called `Ready` in a `Pool`'s `.status.conditions`. By default, an
+interval of 10 seconds is recommended. The pool controller must set
+this condition to `True` if the pool is ready for workload, otherwise
```
friegger: I assume the controller is also responsible to prevent weird combinations like `state` `Error` and condition `Ready` is `True`.

ieps/15-pool-health.md
```diff
@@ -0,0 +1,108 @@
...
+> [!IMPORTANT] The pool controller should not inspect timestamps
+> of the conditions / leases. This is prone to failure due to clock
+> skew. Instead, pool controllers should check if there were updates
+> since the last inspection and only if there were no updates for a
+> specific duration mark the pool as non-ready.
+
```
friegger, line 154: Maybe a small mermaid or ascii diagram of the involved resources and controllers would make it even easier to grasp.

ieps/13-pool-health.md
```diff
@@ -0,0 +1,108 @@
...
+Inside these namespaces, for each `<Type>Pool`, a pool controller
+must acquire and keep a `coordination.k8s.io/Lease`:
+For example, if there is a `MachinePool`
+`my-machinepool`, the controller of this `MachinePool` must acquire and
+keep a lease called `ironcore-machinepool-lease/my-machinepool`. The
+recommended default for setting on the lease is 40 seconds.
+
+Additionally, each pool controller must regularly update a condition
+called `Ready` in a `Pool`'s `.status.conditions`. By default, an
+interval of 10 seconds is recommended. The pool controller must set
```
friegger: These intervals and and grace periods are also used for the K8s node Leases, worth mentioning that these are therefor proven values.

ieps/13-pool-health.md
friegger: It seems at the moment only [MachinePoolStateReady](https://github.com/ironcore-dev/ironcore/blob/78d434af30923f9e5d7c1b7f485ca0dafa2e53ff/poollet/machinepoollet/controllers/machinepool_controller.go#L131) is used.

ieps/15-pool-health.md
adracus: But we don't actually know whether the pool is healthy or not, hence `Unknown` - same pattern the k8s node controller employs. The one that actually can set the condition to `False` is the pool itself.

ieps/15-pool-health.md
adracus: Yes, we're using the 'default' k8s condition type here:

```go
type ConditionStatus string

// These are valid condition statuses. "ConditionTrue" means a resource is in the condition.
// "ConditionFalse" means a resource is not in the condition. "ConditionUnknown" means kubernetes
// can't decide if a resource is in the condition or not. In the future, we could add other
// intermediate conditions, e.g. ConditionDegraded.
const (
	ConditionTrue    ConditionStatus = "True"
	ConditionFalse   ConditionStatus = "False"
	ConditionUnknown ConditionStatus = "Unknown"
)
```

ieps/13-pool-health.md
adracus: Added a deprecation notice for that field. Actually, this field is currently not evaluated (which should be the case for e.g. the scheduler).

ieps/13-pool-health.md
adracus: Added a mention for the k8s defaults.

ieps/15-pool-health.md
adracus: Good idea in general - I considered it but felt the flow is linear enough that prose captures it well without the extra maintenance burden of keeping a diagram in sync. Let me know if anything in the description feels unclear though, happy to improve it.

ieps/15-pool-health.md
friegger: No I meant to make the sentence clearer:

> A pool is deemed
> unresponsive if there was no progress on neither the pool's status nor the pool's lease.

A pool is marked `Unknown`, if the lease has not been renewed in time **and** the condition `Ready` is not true.

adracus: But that's not the case: The `Ready` condition is usually reported by the poollet controller. Now, if the poollet controller becomes unresponsive, the `Ready` condition on the `Pool` resource is stale and might still be `true`. The controller on ironcore-side must then go ahead and overwrite the `Ready` condition of the pool and set it to `Unknown`.

friegger: Ah, now I get it, `no progress on the pool's status` means that the `lastTransitionTime` of the `Ready` condition is older than the interval to update it, correct?

adracus: No progress means that the pool controller has not seen updates of either the `Ready` condition or the lease within an allotted time. We do not compare the timestamps to counteract clock skew issues (see this paragraph):
> 
> [!IMPORTANT] The pool controller should not inspect timestamps
> of the conditions / leases. This is prone to failure due to clock
> skew. Instead, pool controllers should check if there were updates
> since the last inspection and only if there were no updates for a
> specific duration mark the pool as non-ready.

afritzler (approved):

ieps/15-pool-health.md
```diff
@@ -0,0 +1,161 @@
...
+  holderIdentity: machinepoollet-07a5ea9b9b072c4a5f3d1c3702_0c8914f7-0f35-440e-8676-7844977d3a05
+  leaseDurationSeconds: 3600
+  renewTime: "2026-05-01T13:18:48.065888Z"
+```
+
+Additionally, each pool controller must regularly update a condition
+called `Ready` in a `Pool`'s `.status.conditions`. By default, an
+interval of 10 seconds is recommended (taken from Kubernetes' kubelet defaults).
+The pool controller must set this condition to `True` if the pool is ready
+for workload, otherwise it must explicitly set it to `False`.
```
balpert89, line 99: are there any concrete examples *when* the pool health is actually determined as `Unhealthy`? Kubelet defines a well-known set of conditions that leads to a `NotReady` status, I am not aware of such conditions for the `*Pool` statuses.
Should a first set of conditions be defined here as well?

ieps/15-pool-health.md
```diff
@@ -0,0 +1,161 @@
...
+  conditions:
+  - type: Ready
+    status: Unknown
+    reason: NoHeartbeatReceived
+    message: No heartbeat for the machine pool has been received within the past time.
+    observedGeneration: 4
+    lastTransitionTime: "2026-05-01T13:18:48.065888Z"
+```
+
+For each pool type (`MachinePool`, `VolumePool`, `BucketPool`), a controller is
```
balpert89, line 139: I assume this controller is added to the `controller-manager` component. Can be misinterpreted to be part of the `*Poollet` controller set. It is implied by logic, but an explicit statement can help clarity.

ieps/15-pool-health.md
```diff
@@ -0,0 +1,161 @@
...
+If a pool is unresponsive for a specific duration, its `Ready` condition
+is marked as `Unknown`. A recommended default for the
+`<type>pool-monitor-grace-period` is 50 seconds. A pool is deemed
+unresponsive if there was no progress on neither the pool's status
+nor the pool's lease.
+
+> [!IMPORTANT] The pool controller should not inspect timestamps
+> of the conditions / leases. This is prone to failure due to clock
+> skew. Instead, pool controllers should check if there were updates
+> since the last inspection and only if there were no updates for a
```
balpert89, line 152: where is the state of `last inspection` tracked? what happens when `controller-manager` component restarts?

balpert89 (changes requested):

balpert89:
IEP number is off because of intermittent merges. https://github.com/ironcore-dev/enhancements/pull/19 already claimed `15`, I propose `16` here :)

peanball:
> IEP number is off because of intermittent merges. https://github.com/ironcore-dev/enhancements/pull/19 already claimed `15`, I propose `16` here :)

Please just assign the next number in sequence before merging.

See [README.md](https://github.com/ironcore-dev/enhancements).

ieps/15-pool-health.md
adracus: A `*Pool` determines the status of its `Ready` condition currently depending on its own computations (e.g. if all required services are healthy, enough memory / CPU is available).

The aim of this IEP is not to define these computations but to define what happens if a `*Pool` is not responsive (updating the `Ready` condition to `Unknown` if a `*Pool` is not responsive).

For now, such a set of conditions is out of focus for this IEP.

ieps/15-pool-health.md
adracus: Comparable to how the k8s node controller does this: Keep an in-memory cache of the last inspection. On restart, this cache gets populated again, causing a time of `2 * gracePeriod` for unhealthy pools to be marked unhealthy.
Since this proposal focuses on the semantics, I intentionally left the implementation details out if they're not relevant to the proposal.

balpert89 (approved):

PR #7, opened 2026-04-16
Prometheus Metrics for the SONiC Operator and Managed Switches

peanball:
Metrics are collected from the sonic-operator and from each of the switches managed by the operator. Prometheus discovers switches based on information provided by the sonic-operator via HTTP Service Discovery.

Key points:

- On-demand scraping of switch metrics avoids stale data and a single point of failure for metric delivery. Each switch exposes its own Prometheus endpoint.
- Values originate in the Redis databases managed by SONiC and are mapped to Prometheus metrics via a YAML configuration that allows operators to add or remove metrics as needed. A smaller set of code-based metrics (device readiness, interface counts) is always present.
- Proposes a default set of metrics covering operationally relevant performance and health information (interface state, traffic and error counters, SFP transceiver health, chassis temperatures) as a base for alerting.

Alternatives to the proposed approach via standard APIs (gNMI, SNMP) and other collection patterns (operator-as-proxy, Pushgateway) are discussed.

This proposal is primarily aimed at users of switches running the SONiC OS.

### DIFF
--- a/ieps/15-sonic-operator-and-switch-metrics.md
+++ b/ieps/15-sonic-operator-and-switch-metrics.md
@@ -1,0 +1,412 @@
+---
+title: Prometheus Metrics for the SONiC Operator and Managed Switches
+iep-number: 15
+creation-date: 2026-04-16
+status: draft
+authors:
+- "@peanball"
+reviewers:
+- "@maxmoehl"
+- "@byteocean"
+- "@Giluerre"
+---
+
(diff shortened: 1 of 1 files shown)


Giluerre (approved):
Looks good to me. I really like the SD part.

maxmoehl (approved):

hardikdr (approved):

PR #15, opened 2026-04-27
Server Claim Evolution IEP

maxmoehl:


ieps/tbd-evolve-server-claim.md
```diff
@@ -0,0 +1,415 @@
...
+type Toleration struct {
+	Key      string             `json:"key"`
+	Operator TolerationOperator `json:"operator,omitempty"`
+	Value    string             `json:"value,omitempty"`
+	Effect   TaintEffect        `json:"effect,omitempty"`
+}
+```
+
+Well known taint keys are:
+* `metal.ironcore.dev/needs-maintenance`
```
adracus: How about calling this just `metal.ironcore.dev/maintenance`?

ieps/20-evolve-server-claim.md
```diff
@@ -0,0 +1,415 @@
...
+	// SystemUUID is the unique identifier for the server.
+	// +required
+	SystemUUID string `json:"systemUUID"`
+
+	// SystemURI is the unique URI for the server resource in REDFISH API.
+	SystemURI string `json:"systemURI,omitempty"`
+
+	// Power specifies the desired power state of the server.
+	// +optional
+	Power Power `json:"power,omitempty"`
```
adracus, line 129: Should we keep this? Or should we also drop this in favor of the `Power` field on `ServerClaim`?

ieps/tbd-evolve-server-claim.md
```diff
@@ -0,0 +1,415 @@
...
+default, the flag itself will default to `PXE`.
+
+```go
+type BootMethod string
+
+const (
+	BootMethodHTTP = "HTTP"
+	BootMethodPXE = "PXE"
+)
+```
```
adracus: If we might anyways collapse PXE + HTTP booting, should we even introduce this? And should we even 'specify' that we do network boot or rather just call it 'image boot'? We could then (on the `ServerClaim` type) introduce `BootOrder` which is a slice of structs that can be either image boot or device boot, like so:

```go
type ImageBootSource struct {
}

type DeviceBootSource struct {
	Device string `json:"device,omitempty"`
}

type BootSource struct {
	Image *ImageBootSource `json:"image,omitempty"`
	Device *DeviceBootSource `json:"device,omitempty"`
}

type ServerClaimSpec struct {
	...
	BootOrder []BootSource `json:"bootOrder,omitempty"`
	...
}
```

That's just a braindump that came to my mind, also happy to completely scrap this idea.

ieps/tbd-evolve-server-claim.md
```diff
@@ -0,0 +1,415 @@
...
+
+	// Priority can be any value from -100 to 100. The default is 0. A higher
+	// priority claim gets preferred when multiple claims try to claim the same
+	// server. A claim referencing a specific server will always win over a
+	// claim with a selector.
+	// +kubebuilder:validation:Minimum=-100
+	// +kubebuilder:validation:Maximum=100
+	// +kubebuilder:default=0
+	// +optional
+	Priority int `json:"priority,omitempty"`
```
adracus: We should get rid of priority IMO. The only thing being able to prioritize are taints + tolerations that can evict workload and then schedule something on its own afterwards.

ieps/20-evolve-server-claim.md
```diff
@@ -0,0 +1,415 @@
...
+	ServerSelector *metav1.LabelSelector `json:"serverSelector,omitempty"`
+
+	// IgnitionSecretRef is a reference to the Secret object that contains
+	// the ignition configuration for the server.
+	// +optional
+	IgnitionSecretRef *v1.LocalObjectReference `json:"ignitionSecretRef,omitempty"`
+
+	// Image specifies the boot image to be used for the server.
+	// +required
+	Image string `json:"image"`
```
adracus, line 198: I'd actually make this optional. What if e.g. you just want to boot from disk? E.g. if something's already preinstalled on a server or with some SoC that can 'just' boot.

ieps/tbd-evolve-server-claim.md
```diff
@@ -0,0 +1,415 @@
...
+```go
+type ServerClaimStatus struct {
+	// Phase represents the current phase of the server claim.
+	// +optional
+	Phase Phase `json:"phase,omitempty"`
+
+	// ImageURL is the digest-pinned OCI image reference resolved from the
+	// user-provided image. It contains the necessary artifacts for the
+	// specified boot method as layers in the IronCore format.
+	ImageURL string `json:"imageURL,omitempty"`
```
adracus: If we integrated the `boot-operator` into this component, could we also get rid of this field before even introducing it or would that cause too many changes? I'm asking since we anyways need to adapt that component and thus this could be a good moment to integrate it.

ieps/tbd-evolve-server-claim.md
```diff
@@ -0,0 +1,415 @@
...
+`ImageURL` field using the digest of the image to record what the tag resolved
+to when it was checked and ensure the boot process uses the same image. The
+claim transitions to the `Ready` phase and can now be bound. If the image does
+not contain the necessary artifacts, e.g. not an IronCore image or only contains
+PXE artifacts, but claim specifies HTTP, it enters the `Invalid` phase.
+
+To bind a claim to a server, metal-operator checks the taints and tolerations to
+select suitable servers for a claim. Once a match is found the claim is bound to
+the server and the server transitions to the `Reserved` state while the claim
+becomes `Bound`.
```
adracus: Why don't we follow the pod lifecycle more here ( https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase )? We could initially have `Pending` until it's assigned to a `Server` and booted. Once it's booted successfully, it enters `Running` state. If it's failing to boot, it can be in a `CrashLoopBackoff` state similar to pods. I think that's better than letting this implementation detail of the dance between `boot-operator` and `metal-operator` shine through (with the `Ready` phase etc.).

ieps/20-evolve-server-claim.md
```diff
@@ -0,0 +1,415 @@
...
+HTTP booting the metal-operator sets PXE boot as well but the DHCP server must
+be configured to only reply to PXE or HTTP boot requests depending on how the
+claim is configured.
+
+### Onboarding Flow
+
+When onboarding a new server, it is created with its state set to `Initial`.
+When metal-operator first picks up the server it adds the
+`metal.ironcore.dev/initial` taint with the `NoBind` effect to prevent claims
+from getting bound to this server. It also creates a claim which references that
```
adracus, line 287: Why don't we already create it with state `Available` + a well-known `initial` taint directly applied? We could even make it configurable / up to the `metal-operator` what should be initial taints / tolerations (similar to how `kubelet`'s `registerWithTaints` - see https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/ ). Then the initial probing / discovery controller just watches for `Server`s with that specific taint and probe it. We'd have a state less and we'd even be more flexible on which `Server`s we want to discover and which we don't want to discover / probe.

ieps/tbd-evolve-server-claim.md
```diff
@@ -0,0 +1,415 @@
...
+To schedule a maintenance, the `metal.ironcore.dev/needs-maintenance` taint is
+added to a server and a maintenance object referencing the server is created.
+Depending on the effect of the taint, the maintenance is either executed
+gracefully or forcefully, see [taint effects](#taint-effects) for details. In
+the graceful case the information about this taint is propagated to the workload
+manager which has to react to the taint. The workload is migrated off the server
+and eventually the claim is released allowing the server to transition to the
+`Maintenance` state and the maintenance to begin. If, as part of a maintenance,
+the server must be booted a claim with the appropriate toleration must be
+created with the necessary information.
```
adracus: How about we adjust the flow like this (comparable to what you suggested / we talked about offline):
* Maintenance object is created referencing a server to maintain
* Some soft / hard confirmation on the server is requested (e.g. 'yes' annotation on the `Server` / `Maintenance` object).
* Once maintenance is accepted, the maintenance controller adds the taint to the node, evicting workload.
* The maintenance controller can now, per maintenance object to reconcile, 'call' that maintenance's reconciliation functionality. This functionality can then e.g. create a `ServerClaim` if required with its config: For maintenances just requiring the server to be offline, it could just have `Power: Off` and be good. Other maintenance reconciliations could even create multiple `ServerClaim`s until done.
* The `ServerClaim`'s state will not have a special `Maintenance` state, maintenance is just a process happening with specific taints / potentially also an annotation on the server.
* Eventually, once all `Maintenance`s are done, the taints are removed and the server becomes available normally.

ieps/tbd-evolve-server-claim.md
maxmoehl: I'm not sure what you mean by that. The idea here is that, during validation, we have to resolve the digest. When we did that once we may as well cache the result to avoid having to re-do it at boot time when the image is actually requested. It also allows a user to transparently see what exactly was resolved for booting when tracking a tag that is re-tagged.

ieps/20-evolve-server-claim.md
maxmoehl: Especially the ability to configure the taints a server starts out with maps well to the requirement of performing additional checks beyond what metal-operator already does today. We have to be a bit careful how we structure those and what the discovery boot is tolerating to prevent a dead-lock. I will see how we can get this integrated.

ieps/tbd-evolve-server-claim.md
maxmoehl: The entire lifecycle of the claim is managed by metal-operator, including the validation as outlined in the first paragraph.

I didn't follow the pod lifecycle as the lifecycle is different for a claim. A claim isn't "running" and we have no means to detect crashes. The one state I could imaging is `ImagePullBackoff` though even that is somewhat misleading as we don't actually pull the image during validation but just check that it exists.

ieps/tbd-evolve-server-claim.md
maxmoehl: The goal of this is to allow a simple ordering of multiple boots during a maintenance and ordering them reliably without relying on the creation timestamp or something else.

ieps/20-evolve-server-claim.md
maxmoehl: Hmm, I'm very much in favor of dropping this. What I'm now wondering is whether we should include such smaller clean-ups in this proposal or do them separately to keep the focus of this proposal.

ieps/20-evolve-server-claim.md
maxmoehl: This was just taken over from the existing spec. I'm open to relaxing the requirement on this, in that case it would also skip the validation of the image before considering the claim to be ready.

ieps/tbd-evolve-server-claim.md
```diff
@@ -0,0 +1,415 @@
...
+of the new claim phases metal-operator will also start validating the provided
+OCI URL to ensure it's valid.
+
+#### 3. Migrate Discovery Flow
+
+Metal-operator will be adjusted to no longer rely on internal handling via
+server boot configurations to instead use the initial taint and a claim to
+perform the discovery boot.
+
+#### 4. Migrate Maintenance Flow
```
afritzler: I think we should keep the `ServerMaintenance` type rather than reducing it to a passive reference on the `Server` because a `ServerMaintenance` acts as the active orchestrator.

The maintenance lifecycle is fundamentally different from a regular claim lifecycle. A maintenance has intent (firmware update, disk replacement, reprobing, etc.), it has ordering requirements (multiple steps), and it needs to track its own progress. Collapsing this into "just add a taint and create claims" loses the declarative intent of what maintenance is being performed and why.

`ServerMaintenance` should evolve into a orchestrating type:
- It declares the desired maintenance operation(s)
- It carries the scheduling metadata (priority, urgency, allowed windows)
- Once granted (i.e. the system confirms it can proceed), the maintenance controller applies the `metal.ironcore.dev/maintenance` taint to the `Server` and not the other way around

ieps/tbd-evolve-server-claim.md
afritzler: Agree with @adracus on that one. The `Priority` field on a `ServerClaim `feels misplaced. In practice, the only scenario where you need priority-based preemption is maintenance and not regular workload scheduling. Regular claims compete via label selectors and taints/tolerations which is sufficient.

afritzler (changes requested):
I think we should keep the ServerMaintenance type rather than reducing it to a passive reference on the Server. The way I see it, ServerMaintenance acts as the active orchestrator.

The maintenance lifecycle is fundamentally different from a regular claim lifecycle. A maintenance has intent (firmware update, disk replacement, reprobing, etc.), it has ordering requirements (multiple steps), and it needs to track its own progress. Collapsing this into "just add a taint and create claims" loses the declarative intent of what maintenance is being performed and why.

This gives us a clean separation:
  - ServerClaim: "I want to boot this server matching these criteria"
  - ServerMaintenance: "This server needs maintenance with this urgency, once granted I taint it and orchestrate the steps"

ieps/20-evolve-server-claim.md
maxmoehl: WDYT @afritzler?

maxmoehl:
> I think we should keep the ServerMaintenance type rather than reducing it to a passive reference on the Server. The way I see it, ServerMaintenance acts as the active orchestrator.

That was already changed in the proposal, the maintenance stays as it is today.

ieps/tbd-evolve-server-claim.md
maxmoehl: > * Maintenance object is created referencing a server to maintain
> * Some soft / hard confirmation on the server is requested (e.g. 'yes' annotation on the `Server` / `Maintenance` object).

These steps are implied as this is already how it works today to my understanding.

> * Once maintenance is accepted, the maintenance controller adds the taint to the node, evicting workload.

Is this then something that is required by the contract? Or can a maintenance also happen while a claim is bound?

> * The maintenance controller can now, per maintenance object to reconcile, 'call' that maintenance's reconciliation functionality. This functionality can then e.g. create a `ServerClaim` if required with its config: For maintenances just requiring the server to be offline, it could just have `Power: Off` and be good. Other maintenance reconciliations could even create multiple `ServerClaim`s until done.

If it just needs the server to be powered off it could skip the claim entirely as an unclaimed server should always be powered off.

ieps/tbd-evolve-server-claim.md
maxmoehl: The semantics of a server claim are currently to network boot once and after that it depends on the BIOS of the server. The goal of the proposed change is to be more specific about how the specified image is booted. I want to avoid a situation like we have in boot-operator where it is mostly undefined behavior if you have an environment which mixes HTTP and PXE booting.

ieps/tbd-evolve-server-claim.md
adracus: Ok, if it's for identifying what is the *exact* image booted, how about we go for the `imageID` approach like in Kubernetes' pods' status ( https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#containerstatus-v1-core )?

Like in k8s, a Pod has sth like

```
status:
  containerStatuses:
    - name: nginx
      image: nginx:1.25
      imageID: containerd://sha256:3f8a4339aaddc6d7b5c8e9f0b1234567890abcdef1234567890abcdef12345678
      ready: true
      restartCount: 0
```

maxmoehl: I'm happy to change the name, but as we have no image store in our context the content will be a digest-URL pointing to the upstream container registry which is why I named it `ImageURL`.

ieps/20-evolve-server-claim.md
nagadeesh-nagaraja: `ServerClaim` isnt the only controller which is using `Server.Spec.Power`.  How will it be handled in other cases?

should every controller which needs server power control have `Power` spec field? 

I also agree that this should be a separate topic?

ieps/tbd-evolve-server-claim.md
adracus: Ok, then how about `sed s/Running/Booted`? And how about removing the `Ready` phase and put in `Pending`?

ieps/tbd-evolve-server-claim.md
adracus: Agree, makes sense

ieps/tbd-evolve-server-claim.md
```diff
@@ -0,0 +1,400 @@
...
+	// Image specifies the boot image to be used for the server.
+	// +required
+	Image string `json:"image"`
+
+	// Tolerations control whether this claim can be bound to a server that is
+	// tainted in a certain way.
+	Tolerations []Toleration `json:"tolerations,omitempty"`
+
+	// BootMethod configures via which method the server is booted.
+	BootMethod BootMethod `json:"bootMethod,omitempty"`
```
atd9876: How is the BootMethod being applied? are we using boot overrides? If we are using the ServerClaim during maintenance where the server firmware is being updated, the system may reboot numerous times so we would need the boot override to be continuous rather than once.

ieps/tbd-evolve-server-claim.md
```diff
@@ -0,0 +1,400 @@
...
+}
+```
+
+#### BootMethod
+
+New. In addition to the type, metal-operator will have a global flag to set the
+default, the flag itself will default to `PXE`.
+
+```go
+type BootMethod string
```
atd9876: User wants the behaviour

Reserved servers should boot from local disk when powered on
PXE boot should only occur when explicitly requested or during initial provisioning

how do we boot from disk or via the BootOrder in the ServerSpec

how do we handle the first boot being PXE/HTTP to deploy the os to local disk, and subsequent boots being to disk or specifically the boot target created during the installation process?

ieps/20-evolve-server-claim.md
```diff
@@ -0,0 +1,400 @@
...
+* `BootConfigurationRef`
+* `MaintenanceBootConfigurationRef`
+
+#### 6. Remove Deprecated Types
+
+Finally, the now unused types are deleted: 
+
+* `ServerBootConfiguration`
+* `HTTPBootConfig`
+* `IPXEBootConfig`
```
atd9876, line 361: # Suspension vs Eviction for Maintenance

## Problem

The IEP proposes `Evict` taint effect to remove ServerClaims during maintenance. This works for stateless servers but is destructive for stateful workloads.

## Eviction Breaks Stateful Servers# Suspension vs Eviction for Maintenance

## Problem

The IEP proposes `Evict` taint effect to remove ServerClaims during maintenance. This works for stateless servers but is destructive for stateful workloads.

## Eviction Breaks Stateful Servers

When a claim is evicted (deleted):

1. **Binding lost** — nothing reserves the server for the original consumer
2. **Server exposed** — another consumer could claim it post-maintenance
3. **Data at risk** — no claim = no signal that disks must be preserved (a no-wipe taint could exist, but relies on it being applied correctly)
4. **Race window** — between maintenance completion and re-claim, server is unprotected and could be claimed by another consumer

## Proposal: Add Suspension

Suspension preserves the claim through maintenance:

- Claim exists + suspended → server stays Reserved → no reassignment possible
- Data preservation is implied (binding persists)
- Post-maintenance: unsuspend → resume on same server, no reprovisioning

### Kubernetes Precedent

| Precedent | Mechanism |
|---|---|
| `Job.spec.suspend` (GA 1.24) | Terminates pods, preserves job + progress, resumes on unsuspend |
| `PV.spec.persistentVolumeReclaimPolicy: Retain` | Preserves data when binding is released |
| Node cordon | Marks unschedulable, object persists, existing work continues |

No generic suspension primitive exists, but bespoke `suspend` fields are idiomatic.

### Proposed API

A claim-level maintenance policy:

```yaml
kind: ServerClaim
spec:
  maintenancePolicy: Suspend  # preserve binding (stateful)
  # or
  maintenancePolicy: Evict    # delete claim (stateless, default)
```

The flow:
1. Server is tainted `Maintenance` (trigger)
2. Controller reads the bound claim's `maintenancePolicy`
3. `Evict` → delete claim, server becomes Available after maintenance
4. `Suspend` → preserve claim, stop workload, maintain, unsuspend, resume

No tolerations needed — the taint lives on the Server (trigger), the policy lives on the claim (intent). One field, one decision.

## When Each Model Applies

| | Evict | Suspend |
|---|---|---|
| Server type | Cattle — identical, replaceable | Pet — specific hardware, local data |
| Reprovision cost | Minutes (image + ignition) | Hours/days (TBs of data, state rebuild) |
| Post-maintenance | New claim, full reprovision | Unsuspend, resume as-is |
| Binding during maintenance | Gone | Preserved |

## Summary

Eviction models servers as ephemeral Pods. Suspension models them as durable PersistentVolumes. Bare metal is often chosen precisely because workloads need local state, specific hardware, or accumulated data — making the server itself stateful. For these, maintenance is a temporary interruption, not a termination of the relationship. Suspension is simpler (no re-claim logic), safer (no unprotected window), and more accurate.


When a claim is evicted (deleted):

1. **Binding lost** — nothing reserves the server for the original consumer
2. **Server exposed** — another consumer could claim it during maintenance
3. **Data at risk** — no claim = no signal that disks must be preserved
4. **Re-claim complexity** — consumer must target the exact same server post-maintenance and signal "don't reprovision"
5. **Race window** — between maintenance completion and re-claim, server is unprotected

## Proposal: Add Suspension

Suspension preserves the claim through maintenance:

- Claim exists + suspended → server stays Reserved → no reassignment possible
- Data preservation is implied (binding persists)
- Post-maintenance: unsuspend → resume on same server, no reprovisioning

### Kubernetes Precedent

| Precedent | Mechanism |
|---|---|
| `Job.spec.suspend` (GA 1.24) | Terminates pods, preserves job + progress, resumes on unsuspend |
| `PV.spec.persistentVolumeReclaimPolicy: Retain` | Preserves data when binding is released |
| Node cordon | Marks unschedulable, object persists, existing work continues |

No generic suspension primitive exists, but bespoke `suspend` fields are idiomatic.

### Proposed API

A claim-level maintenance policy:

```yaml
kind: ServerClaim
spec:
  maintenancePolicy: Suspend  # preserve binding (stateful)
  # or
  maintenancePolicy: Evict    # delete claim (stateless, default)
```

The flow:
1. Server is tainted `Maintenance` (trigger)
2. Controller reads the bound claim's `maintenancePolicy`
3. `Evict` → delete claim, server becomes Available after maintenance
4. `Suspend` → preserve claim, stop workload, maintain, unsuspend, resume

No tolerations needed — the taint lives on the Server (trigger), the policy lives on the claim (intent). One field, one decision.

## When Each Model Applies

| | Evict | Suspend |
|---|---|---|
| Server type | Cattle — identical, replaceable | Pet — specific hardware, local data |
| Reprovision cost | Minutes (image + ignition) | Hours/days (TBs of data, state rebuild) |
| Post-maintenance | New claim, full reprovision | Unsuspend, resume as-is |
| Binding during maintenance | Gone | Preserved |

## Summary

Eviction models servers as ephemeral Pods. Suspension models them as durable PersistentVolumes. Bare metal is often chosen precisely because workloads need local state, specific hardware, or accumulated data — making the server itself stateful. For these, maintenance is a temporary interruption, not a termination of the relationship. Suspension is simpler (no re-claim logic), safer (no unprotected window), and more accurate.

ieps/20-evolve-server-claim.md
```diff
@@ -0,0 +1,400 @@
...
+is set to `Available` allowing the server to be claimed by any claim.
+
+### Maintenance Flow
+
+In addition to the existing maintenance flow where a maintenance gets bound to a
+server object, the maintenance controller may choose to assign a taint to the
+server to evict any current claim or prevent new claims from binding to the
+server. The `metal.ironcore.dev/maintenance` taint key is well-known for this
+kind of taint. Once the maintenance is done, the maintenance controller clears
+any taints it previously set and the server can be used as before.
```
atd9876, line 301: # Suspension vs Eviction for Maintenance

## Problem

The IEP proposes `Evict` taint effect to remove ServerClaims during maintenance. This works for stateless servers but is destructive for stateful workloads.

## Eviction Breaks Stateful Servers

When a claim is evicted (deleted):

1. **Server unprotected** — no binding reserves the server, so another consumer could claim it before the original consumer re-claims
2. **Data at risk** — no claim = no signal that disks must be preserved (a no-wipe taint could exist, but relies on it being applied correctly)

## Proposal: Add Suspension

Suspension preserves the claim through maintenance:

- Claim exists + suspended → server stays Reserved → no reassignment possible
- Data preservation is implied (binding persists)
- Post-maintenance: unsuspend → resume on same server, no reprovisioning

### Kubernetes Precedent

| Precedent | Mechanism |
|---|---|
| `Job.spec.suspend` (GA 1.24) | Terminates pods, preserves job + progress, resumes on unsuspend |
| `PV.spec.persistentVolumeReclaimPolicy: Retain` | Preserves data when binding is released |
| Node cordon | Marks unschedulable, object persists, existing work continues |

No generic suspension primitive exists, but bespoke `suspend` fields are idiomatic.

### Proposed API

A claim-level maintenance policy:

```yaml
kind: ServerClaim
spec:
  maintenancePolicy: Suspend  # preserve binding (stateful)
  # or
  maintenancePolicy: Evict    # delete claim (stateless, default)
```

The flow:
1. Server is tainted `Maintenance` (trigger)
2. Controller reads the bound claim's `maintenancePolicy`
3. `Evict` → delete claim, server becomes Available after maintenance
4. `Suspend` → preserve claim, stop workload, maintain, unsuspend, resume

No tolerations needed — the taint lives on the Server (trigger), the policy lives on the claim (intent). One field, one decision.

## When Each Model Applies

| | Evict | Suspend |
|---|---|---|
| Server type | Cattle — identical, replaceable | Pet — specific hardware, local data |
| Reprovision cost | Minutes (image + ignition) | Hours/days (TBs of data, state rebuild) |
| Post-maintenance | New claim, full reprovision | Unsuspend, resume as-is |
| Binding during maintenance | Gone | Preserved |

## Summary

Eviction models servers as ephemeral Pods. Suspension models them as durable PersistentVolumes. Bare metal is often chosen precisely because workloads need local state, specific hardware, or accumulated data — making the server itself stateful. For these, maintenance is a temporary interruption, not a termination of the relationship. Suspension is simpler (no re-claim logic), safer (no unprotected window), and more accurate.

maxmoehl: Do I get it right that you want to preserve the ability to have stateful workloads that are not evicted by maintenance but rather just "paused"? If so, I'm asking you to please get to the point right away in your review comments and skip the waffle.

In our scenario servers are cattle, we do not want pets. Can you elaborate a bit why you need stateful servers that persist some state across maintenances and what that state is?

ieps/20-evolve-server-claim.md
maxmoehl: I'm opting to keep this for now. Smaller clean ups can be done via issues and I want to focus on the API changes proposed.

ieps/tbd-evolve-server-claim.md
maxmoehl: This only handles the first boot like it is today. metal-operator never cared about how a server is booted in general, only when a claim is bound for the first time does it intervene via the network boot override. All this does is specify per-claim whether it's HTTP or PXE. Other than that the behavior remains as it is today.

ieps/tbd-evolve-server-claim.md
maxmoehl: See other comment.

ieps/tbd-evolve-server-claim.md
maxmoehl: We decided to drop the boot method for now and follow the pod lifecycle more closely, e.g. surfacing a condition in case of an invalid image.

ieps/tbd-evolve-server-claim.md
maxmoehl: After some in-person discussion the states were re-worked. Please have another look.

ieps/tbd-evolve-server-claim.md
maxmoehl: We ended up dropping this field in the light of upcoming proposals.

adracus (approved):
Very nice IEP 🚀

ieps/20-evolve-server-claim.md
atd9876: I agree, in homogeneous single use cluster deployments this does not arise.

ieps/tbd-evolve-server-claim.md
```diff
@@ -0,0 +1,361 @@
+---
+title: Evolve ServerClaim into the Universal Boot Primitive
+
+iep-number: tbd
```
afritzler: We probably want to add a number here.

ieps/20-evolve-server-claim.md
```diff
@@ -0,0 +1,361 @@
...
+* `metal.ironcore.dev/initial`
+
+#### Server
+
+Add:
+* `Taints`
+
+Remove:
+* `BootConfigurationRef`
+* `MaintenanceBootConfigurationRef`
```
afritzler, line 116: Do you mean `serverMaintenanceRef` here? If so why is it still in the `ServerSpec` below?

ieps/20-evolve-server-claim.md
afritzler: So we keep the `Initial` state for now and factor this out later as I am with @adracus on that one.

afritzler (changes requested):

ieps/tbd-evolve-server-claim.md
maxmoehl: Yes, will do that just before the merge.

ieps/20-evolve-server-claim.md
maxmoehl: No, this was only about getting rid of ServerBootConfiguration. Maintenance stays.

afritzler (approved):

PR #16, opened 2026-04-27
IronCore Image V2 proposal

adracus:
# Proposal

Our current image specification has evolved a lot since the beginning.
This proposal tries to cleanly describe how we boot and to enhance upon the ideas we have so far.

### DIFF
--- a/ieps/tbd-ironcore-image-v2.md
+++ b/ieps/tbd-ironcore-image-v2.md
@@ -49,6 +49,7 @@ different ways to boot:
 * [UKI](https://github.com/ironcore-dev/ironcore-image/blob/f89db68d5e9323ed5d6024ba70897e077504c25b/image.go#L21)
 * [ISO](https://github.com/ironcore-dev/ironcore-image/blob/f89db68d5e9323ed5d6024ba70897e077504c25b/image.go#L22)
 * [Direct kernel + initramfs](https://github.com/ironcore-dev/ironcore-image/blob/f89db68d5e9323ed5d6024ba70897e077504c25b/image.go#L19)
+* [Squashfs](https://github.com/ironcore-dev/ironcore-image/blob/main/image.go#L20) 
 
 There's even a 'missing' way to boot: 'Raw' EFI executables.
 
@@ -78,10 +79,10 @@ functionality using e.g. initramfs layering.
 * Create the following cleanly separated boot modes, each having their own
   config media and manifest type:
   * UKI-like boot mode
-  * ISO boot mode
-  * Raw EFI boot mode
-* Don't recommend specifying UKIs as UKI binary blobs anymore. Instead,
-  recommend specifying them as UKI image with the distinct layers,
+  * Disk boot mode
+  * Raw EFI executable boot mode
+* Don't specify UKIs as UKI binary blobs anymore. Instead,
+  specify their components (kernel, initrd, stub etc.) as distinct layers,
   allowing for layer deduplication. If someone still chooses to e.g. specify
   a UKI as raw EFI executable, no deduping / further processing will happen.
 * Create facilities for installer images: Namely, composability of initramfs
@@ -198,13 +199,13 @@ Disk booting specifies a disk to boot from.
 
 #### Disk
 
-The disk to boot. Required.
+The disk image to boot.
 
 **Media Type**
 
-Currently, only ISOs via
+Currently, only `raw` files (that could be directly `dd`ed on disk) via
 
-`application/vnd.ironcore.iso`
+`application/vnd.ironcore.raw`
 
 ## Alternatives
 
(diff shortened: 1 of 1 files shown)


ieps/tbd-ironcore-image-v2.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Goals
+
+* Create the following cleanly separated boot modes, each having their own
+  config media and manifest type:
+  * UKI-like boot mode
+  * ISO boot mode
+  * Raw EFI boot mode
+* Don't allow specifying UKIs as entire UKI binaries anymore. Instead,
+  split them into their components, allowing layer deduplication.
```
maxmoehl: Well, you can't really forbid that: an opaque UKI boot (where none of the components know about the internals of the UKI) would become the raw EFI boot.

ieps/tbd-ironcore-image-v2.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Non-Goals
+
+* None identified yet
+
+## Proposal
+
+Create the following boot types and specs:
+
+### UKI booting
```
maxmoehl: Naming this after UKI seems a bit off for me. Essentially you are providing the necessary artifacts required to boot linux in a decomposed way. No one would stop you from serving them via iPXE instead of combining them to a UKI. I would prefer a naming that focuses on the things it contains, not the use-case we have in mind for it.

What I don't know: is this format specific to linux or could you boot a BSD using the same components? If it's generic I'd use something like `application/vnd.ironcore.os`, if not maybe `application/vnd.ironcore.linux`?

maxmoehl (commented):
I'd structure all the media types as `$purpose+$format`. For a kernel the type would then be:

```
application/vnd.ironcore.kernel+efi
```

And a UKI could be:
```
application/vnd.ironcore.uki+efi
```

It gets a bit weird for the raw EFI binaries, but maybe:

```
application/vnd.ironcore.raw+efi
```

For the initrd its nested so that would then be:

```
# uncompressed
application/vnd.ironcore.initrd+cpio
# gzip compressed
application/vnd.ironcore.initrd.cpio+gzip
# ...
```

ieps/tbd-ironcore-image-v2.md
adracus: Reworded the section to be a recommendation. One can definitely choose to use the 'raw' mode but won't get the benefits of deduplication / any other post processing we might offer.

ieps/tbd-ironcore-image-v2.md
adracus: I also struggled a bit with the naming: Although it's not a UKI, its the components that make up a UKI. Creating another name for it like e.g. `bootkit` or sth. else requires additional explanation to anyone wanting to use it. If we keep `uki` (although technically not entirely accurate) we immediately set the right mental model for anybody to know how it's comprised. I'd actually still opt for `uki` as a name since it's shorter and sets up the right model to think about it.

adracus:
Regarding the naming with `$purpose+$format` - I deliberately chose to name it like e.g. `application/vnd.ironcore.initrd.cpio` (base) / `application/vnd.ironcore.initrd.cpio+xz` as it closely follows the style used for the container OCI specification ( https://github.com/opencontainers/image-spec/blob/main/media-types.md ):

In that spec, for example `tar` layers are also written as `application/vnd.oci.image.layer.v1.tar` and the `+` suffix is only used to disambiguate within that format, e.g. `application/vnd.oci.image.layer.v1.tar+gzip`

maxmoehl:
> In that spec, for example `tar` layers are also written as `application/vnd.oci.image.layer.v1.tar` and the `+` suffix is only used to disambiguate within that format, e.g. `application/vnd.oci.image.layer.v1.tar+gzip`

And for JSON it's always with a `+`, this seems completely arbitrary to me. Following what's already there is always preferred so I guess we stick with the arbitrary system and try to reasonably model it to our use-case. If someone figures out what the system is, I'd love to learn.

ieps/tbd-ironcore-image-v2.md
maxmoehl: Well, my argument is rather that it's not _just_ for a UKI. A UKI contains all necessary artifacts to boot a Linux OS, but those are the same components you'd need for (i)PXE, heck even for the ISO boot you need those same components. It's all just different ways of packing them together.

I'd like to avoid a situation like we currently have where we call the layer a UKI because that's what we had in mind when in reality it's just an arbitrary EFI binary.

adracus:
The `+json` suffix is still only used for disambiguation inside a type: For example, take a look at `vnd.oci.image.config.v1+json`: The type is a config, but to disambiguate (it could be represented either in e.g. yaml or toml) they use the `+json` suffix. For other parts where a specific file format is required, like `application/vnd.oci.image.layer.v1.tar` the `.tar` says that it must be `.tar` and disambiguation is just e.g. `+xz` if compressed.

Another project that publishes OCI artifacts is Helm that also follows the convention I mentioned ( https://helm.sh/blog/helm-oci-mediatypes/ ). I'd recommend sticking with what is already known throughout the container OCI spec and surrounding projects already do.

adracus:
Superseded by #49 (due to change to fork-based contribution model)

PR #17, opened 2026-04-27
Introducing ServerReadinessRule to ensure Server readiness

### LINKED ISSUE #11: Availability gating for IronCore servers
## Summary

Introduce a generic availability gating mechanism that allows external services to block servers from reaching `Available` state until their checks pass. The first use case is network topology verification, but the mechanism is designed to support any type of check.

## Motivation

Servers entering the fleet may be physically wired incorrectly or fail other pre-availability checks. Without a gating mechanism, a server can reach `Available` state and get claimed before these checks pass, leading to issues at workload runtime. The gating should be automatic for new and maintenance-returning servers, transparent to the workload layer, and extensible to multiple check types.

### New state: `Manageable`

A new server state is introduced between `Discovery` and `Available`, borrowed from OpenStack Ironic:

```mermaid
stateDiagram-v2
    Discovery --> Manageable: done, taints present
    Discovery --> Available: done, no taints
    Manageable --> Available: all taints removed
    Available --> Manageable: taint added
    Maintenance --> Manageable: exit, taints present
    Maintenance --> Available: exit, no taints
```

### Opt-in via BMC label

External services register a check by setting a label on the `BMC` object:

```yaml
metadata:
  labels:
    metal.ironcore.dev/check-network: "true"
```

The label key format is `metal.ironcore.dev/check-<topic>: "true"`. Multiple labels can be present simultaneously.

The `BMC` object is the natural carrier for these labels because it is the first object created in both onboarding paths:
- **argora-managed servers**: argora creates the BMC directly from NetBox data (inline IP, no `Endpoint` object is created)
- **auto-discovered servers**: metal-operator's `EndpointReconciler` creates the BMC from a discovered `Endpoint` object

In both cases the BMC exists before the server reaches `Initial` state, making it the earliest reliable point to read check requirements.

### Taint on Server

At the `Initial` → `Discovery` transition, metal-operator reads the `BMC` labels and sets a corresponding taint on the `Server` for each matching label:

```yaml
spec:
  taints:
  - key: metal.ironcore.dev/check-network
    effect: NoSchedule
```

The taint key mirrors the label key exactly. Reading labels at this transition (rather than at `Initial`) gives external services time to set their labels on the BMC before taints are created.

### Components

**metal-operator**
- Set `check-*` taints at `Initial` → `Discovery` transition based on `BMC` labels
- Implement `Manageable` state in the server state machine
- Write LLDP/interface data to `ServerMetadata` during Discovery

**argora**
- Creates BMC objects with inline IP sourced from NetBox (no `Endpoint` object)
- Continuously reconcile `ServerMetadata.spec.interfaces` with expected topology from NetBox

**metal-gate** (new microservice)
- Watch `BMC` objects and patch `metal.ironcore.dev/check-*` labels based on cluster spec (which checks are required for that server)
- Watch servers in `Manageable` state with matching taints
- Run configured checks, remove taints on pass, alert on mismatch

### ServerMetadata ownership

`ServerMetadata` is proposed as a shared CRD with split ownership by subresource:

- `status.interfaces` — owned by metal-operator, populated with discovered LLDP data during Discovery. metal-operator is the sole writer via the `/status` subresource.
- `spec.interfaces` — owned by external services (e.g. argora), populated with expected topology from an authoritative source (e.g. NetBox). External services write this via normal `Update` on the spec.

This pattern is valid in Kubernetes — it is common for operators to own `status` while external actors drive `spec`. However it changes the original intent of `ServerMetadata` as a metal-operator-internal persistence store, and introduces a coupling between metal-operator and the shape of external data.

See open question 1 below.

## Open Questions

1. Should `ServerMetadata` carry both spec and status, or should a separate argora-owned CRD hold the expected topology?

## Related

- ironcore-dev/metal-operator#762 — ServerMetadata CRD
- ironcore-dev/metal-operator#672 — Server taints

xkonni:
Adds IEP-0014 proposing a generic availability gating mechanism for IronCore servers.

Closes #11

### DIFF
--- a/ieps/14-availability-gating.md
+++ b/ieps/14-availability-gating.md
@@ -27,24 +27,23 @@ reviewers:
     - [Non-Goals](#non-goals)
 - [Proposal](#proposal)
     - [Overview](#overview)
-    - [Bootstrap taint](#bootstrap-taint)
-    - [External gate controllers](#external-gate-controllers)
+    - [ReadinessGateController](#readinessgatecontroller)
     - [ServerReadinessRule CRD](#serverreadinessrule-crd)
-    - [Lifecycle example](#lifecycle-example)
-    - [Re-gating](#re-gating)
+    - [Enforcement modes](#enforcement-modes)
+    - [Lifecycle examples](#lifecycle-examples)
     - [Observability](#observability)
 - [Dependencies](#dependencies)
 - [Related](#related)
 
 ## Summary
 
-Introduce a generic availability gating mechanism that prevents servers from being claimed until all declared readiness gates pass. The design follows the upstream [Kubernetes Node Readiness Controller](https://kubernetes.io/blog/2026/02/03/introducing-node-readiness-controller/) pattern: metal-operator sets a bootstrap taint at server creation, external gate controllers independently manage their own gate-specific taints and report conditions on the `Server` object, and metal-operator removes the bootstrap taint once all gate taints are cleared.
+Introduce a generic availability gating mechanism that prevents servers from being claimed until all declared readiness gates pass. A `ServerReadinessRule` declares which servers to target, which condition to monitor, and which taint to manage. The `ReadinessGateController` inside metal-operator watches rules and `Server` conditions and manages taint add/remove. Any entity with the appropriate permissions can write conditions on `Server` objects — they need not be aware that a rule exists.
 
 ## Motivation
 
 Servers entering the fleet may fail pre-availability checks (e.g. physical wiring, unclean disks). Without a gating mechanism, a server can reach `Available` state and get claimed before these checks pass, leading to issues at workload runtime.
 
-The driving use case is network topology verification: an external controller compares discovered interface data against an authoritative source and only clears its gate condition once the topology matches. Other gate types follow the same pattern without requiring changes to metal-operator.
+The driving use cases are network topology verification (an external controller compares discovered interface data against an authoritative source) and disk sanitization (prior workload data must be wiped before the server is re-claimable). Other gate types follow the same pattern without requiring changes to metal-operator.
 
 ### Goals
 
@@ -56,7 +55,7 @@ The driving use case is network topology verification: an external controller co
 
 ### Non-Goals
 
-- Defining the logic run by individual gate controllers
+- Defining the logic run by individual external controllers
 - Replacing existing maintenance workflows
 - Automatic timeout or force-removal of taints for stuck gates — operators should alert on long-gated servers via the provided metrics
 - Eviction of already-reserved servers based on gate state
@@ -65,80 +64,73 @@ The driving use case is network topology verification: an external controller co
 
 ### Overview
 
-Gating is enforced via `NoClaim` taints on the `Server` object. A server can only be claimed when all `NoClaim` taints are gone. There are two categories of taint:
+The taints of `Server` objects can be manipulated via `ServerReadinessRule` objects. These rules monitor the `.status.conditions` of a `Server` for a specific condition type and status.
 
-- **Bootstrap taint** (`metal.ironcore.dev/not-ready`) — set by metal-operator at trigger points, removed by metal-operator once no other `NoClaim` taints remain
-- **Gate taints** (e.g. `metal.ironcore.dev/network-not-ready`) — set and removed by external gate controllers, one taint key per gate type
+If the specific condition type is not present in the expected status, a taint specified in the rule is applied to the `Server`.
 
-Each controller has exclusive ownership of its taint key. They never interact with each other's taints.
+Once the condition is present in the expected status, the taint is removed.
 
-### Bootstrap taint
+This makes a `Server` unclaimable to `ServerClaim`s that don't have the required tolerations.
 
-Metal-operator sets the bootstrap taint automatically — no feature flag is required. At each trigger point, metal-operator checks whether any `ServerReadinessRule` matches the server. If at least one does, the bootstrap taint is set. Servers with no matching rules are unaffected and proceed normally.
+### ReadinessGateController
 
-| Trigger | Behavior |
-|---|---|
-| `creation` | Set `metal.ironcore.dev/not-ready: NoClaim` at `Initial` → `Discovery` |
-| `maintenance` | Re-set the taint on `Maintenance` exit |
-| `release` | Re-set the taint when a server transitions from `Reserved` back to `Available` |
+The `ReadinessGateController` runs inside metal-operator and:
 
-The bootstrap taint carries the trigger reason as its `value` field:
+- Watches all `ServerReadinessRule` objects and all `Server` objects
+- At each trigger point, evaluates which rules match a server and applies the corresponding taints
+- Watches `Server.status.conditions` and removes a taint when its condition is satisfied
+- Manages rule lifecycle: adds a finalizer at rule creation, removes orphaned taints on rule deletion or selector change
 
-```yaml
-taints:
-  - key: metal.ironcore.dev/not-ready
-    value: "creation"   # or: "maintenance", "release"
-    effect: NoClaim
-```
-
-Gate controllers can inspect the taint value to fast-forward checks where appropriate. A disk sanitization controller seeing `value: "creation"` knows no prior workload ran on the server and can immediately mark its condition `True`. Seeing `value: "release"` it performs the full sanitization. This is a hint — gate controllers are not required to act differently per trigger value.
-
-Metal-operator removes the bootstrap taint once no other `NoClaim` taints remain on the server. The expected flow is that external gate controllers apply their gate taints early in Discovery (which takes several minutes), so by the time Discovery completes the gate taints are in place. If no gate controllers target the server, the bootstrap taint is removed at the end of Discovery and the server proceeds normally.
-
-**`NoClaim` taints do not affect `Reserved` servers.** The taint is only evaluated at claim time — it prevents a new `ServerClaim` from binding but does not evict an existing reservation. Taints added while a server is `Reserved` (e.g. hardware degradation detected during a workload's lifetime) are preserved and not cleaned up on reservation. When the server is released and transitions back to `Available`, the `release` trigger fires: metal-operator re-sets the bootstrap taint and all matching gate controllers re-evaluate. The server becomes claimable again only after all taints are cleared.
-
-### External gate controllers
-
-External gate controllers are independent components, one per gate type. Each gate controller:
-
-- Watches `Server` objects via a label selector defined in the matching `ServerReadinessRule`
-- Applies its own gate taint to matching servers
-- Runs its check and writes its condition to `Server.status.conditions`
-- Removes its own gate taint when its condition passes
-
-Gate controllers do not interact with the bootstrap taint or with each other.
-
-**Re-arm contract:** when a gate controller re-arms at a trigger point (i.e. it applies its gate taint because the bootstrap taint appeared on a matching server), it must reset its condition to `False` before running its check. This ensures a passing condition from a previous cycle does not carry over and cause the gate to be skipped.
-
-Servers are brought into scope via the `spec.serverSelector` in the `ServerReadinessRule` — this is the gate controller's targeting configuration. Any labels already present on the server (hardware type, region, zone, etc.) can be used; no dedicated gate labels are required.
+Any entity with the appropriate permissions can write conditions on `Server.status.conditions`. They do not need to apply or remove taints and need not be aware of any `ServerReadinessRule` that references their condition type.
 
 ### ServerReadinessRule CRD
 
-A `ServerReadinessRule` is a cluster-scoped resource, created once by whoever deploys a gate controller. It declares which servers the gate controller targets, which condition it monitors, and which taint it manages. Gate controllers read rules to determine which servers to target and which taint to manage. Metal-operator watches rules solely for lifecycle management: it adds a finalizer at creation and removes orphaned gate taints on deletion or selector change.
+A `ServerReadinessRule` is a cluster-scoped resource that declares:
 
-**Rule lifecycle managed by metal-operator:**
-
-- At creation, metal-operator adds `metal.ironcore.dev/serverreadinessgc` as a finalizer.
-- On deletion (`deletionTimestamp` set), metal-operator removes the rule's taint from all servers that carry it, then releases the finalizer. The gate controller being unavailable at deletion time does not block cleanup.
-- On `spec.serverSelector` change, metal-operator removes the rule's taint from servers that no longer match the updated selector.
-
-This gives a clean separation: gate controllers own gating logic; metal-operator owns rule lifecycle and taint GC.
+- Which servers to target (`spec.serverSelector`)
+- Which condition to monitor (`spec.condition.type`)
+- Which taint to manage (`spec.taint`)
+- When to enforce the rule (`spec.enforcement`)
 
 ```yaml
 apiVersion: metal.ironcore.dev/v1alpha1
 kind: ServerReadinessRule
 metadata:
-  name: network-readiness
+  name: disk-sanitization
   finalizers:
     - metal.ironcore.dev/serverreadinessgc
 spec:
   serverSelector:
     matchLabels:
       example.com/hardware-type: server-gen2
-      topology.kubernetes.io/region: region-a
+  enforcement:
+    mode: bootstrap
+    triggers:
+      - creation
+      - maintenance
+  condition:
+    type: DiskSanitized
+  taint:
+    key: metal.ironcore.dev/disk-not-ready
+    effect: NoClaim
+---
+apiVersion: metal.ironcore.dev/v1alpha1
+kind: ServerReadinessRule
+metadata:
+  name: network-topology
+  finalizers:
+    - metal.ironcore.dev/serverreadinessgc
+spec:
+  serverSelector:
+    matchLabels:
+      example.com/hardware-type: server-gen2
+  enforcement:
+    mode: bootstrap
+    triggers:
+      - creation
+      - maintenance
   condition:
     type: NetworkTopologyVerified
-    maxAge: 4h
   taint:
     key: metal.ironcore.dev/network-not-ready
     effect: NoClaim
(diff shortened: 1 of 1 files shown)


ieps/14-availability-gating.md
```diff
@@ -0,0 +1,164 @@
...
+At the `Initial` → `Discovery` transition, metal-operator reads the `BMC` labels and sets a corresponding taint on the `Server` for each matching label:
+
+```yaml
+spec:
+  taints:
+  - key: metal.ironcore.dev/gate-network
+    effect: NoBind
+```
+
+The taint key mirrors the label key exactly. Reading labels at this transition (rather than at `Initial`) gives external services time to set their labels on the BMC before taints are created.
```
afritzler: The proposal relies on a timing contract between argora (or any external service) labeling the BMC and metal-operator reading those labels at Initial -> Discovery. This is a race by design. If argora is slow, crashes, or the label reconciliation loop hasn't fired yet, the Server transitions ungated. Kubernetes explicitly solved this problem by having the registering agent itself (kubelet) set the blocking taint at creation time, making it impossible for a resource to exist in a schedulable state before verification. The current proposal inverts that: the Server is born "open" and must be closed by an external signal arriving in time.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,164 @@
...
+```yaml
+metadata:
+  labels:
+    metal.ironcore.dev/gate-network: "true"
+```
+
+The label key format is `metal.ironcore.dev/gate-<topic>: "true"`. Multiple labels can be present simultaneously.
+
+The `BMC` object is the natural carrier for these labels because it is the first object created in both onboarding paths:
+- **argora-managed servers**: argora creates the BMC directly from NetBox data (inline IP, no `Endpoint` object is created)
```
afritzler: I would phrase this a bit more generic. Instead of argora or any other tool I would just refer to 3rd party/external tooling.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Taint on Server
+
+At the `Initial` → `Discovery` transition, metal-operator reads the `BMC` labels and sets a corresponding taint on the `Server` for each matching label:
+
+```yaml
+spec:
+  taints:
+  - key: metal.ironcore.dev/gate-network
+    effect: NoBind
```
afritzler: Can we use Server conditions for the gate status instead of relying solely on taints? In K8s the NodeLifecycleController bridges conditions like Ready or NetworkUnavailable to taints. The condition tells you why, the taint enforces the scheduling effect. I would do the same here: gate controllers report their status via conditions (NetworkTopologyVerified: False), and a gating controller translates unmet conditions into NoBind taints. That gives you observability for free (kubectl describe server shows what's pending) without having to reverse-map taint keys to owning controllers. Wdyt?

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,164 @@
...
+
+#### Option A — Gate at state (`Manageable` state)
+
+The state machine enforces the gate: `Discovery → Manageable` (if taints present) → `Available` (all taints cleared). `Available` only contains genuinely ready servers.
+
+```mermaid
+stateDiagram-v2
+    Discovery --> Manageable: done, taints present
+    Discovery --> Available: done, no taints
+    Manageable --> Available: all taints removed
```
afritzler: I would not introduce a new Manageable state. IMO the state machine should reflect lifecycle stages (Discovery, Available, Reserved, etc.) Not whether external checks passed. Can we instead just have metal-operator set an `uninitialized:NoBind` taint unconditionally at Server creation (like kubelet does with node.cloudprovider.kubernetes.io/uninitialized:NoSchedule) and let gate controllers remove it? That way Available still means "fully configured and ready" because the taint blocks binding regardless of state. Keeps the state machine simple, no new transitions needed, and the concern of "is this server verified" lives entirely in taints/conditions where it belongs.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,164 @@
...
+Servers entering the fleet may be physically wired incorrectly or fail other pre-availability gates. Without a gating mechanism, a server can reach `Available` state and get claimed before these gates pass, leading to issues at workload runtime. The gating should be automatic for new and maintenance-returning servers, transparent to the workload layer, and extensible to multiple gate types.
+
+### Goals
+
+- Block servers from becoming `Available` until all registered gates pass *(Option A; under Option B servers reach `Available` but cannot be claimed)*
+- Support multiple concurrent gate types
+- Work for both argora-managed and auto-discovered servers
+- Be transparent to the workload scheduling layer
+
+### Non-Goals
```
afritzler, line 52: What happens if a server's network topology changes after it reached Available? Can metal-gate re-apply taints or do we only gate on initial discovery? In K8s the NodeLifecycleController continuously monitors conditions and re-taints nodes when they degrade. I would at least consider supporting re-gating here. Otherwise a miscabled server that was fine at discovery time stays bindable even after the topology drifts. Might be out of scope for this IEP but worth calling out explicitly in non-goals if so.

afritzler (changes requested):
Overall I think the gating mechanism makes sense but I would challenge a few design decisions:

- The BMC label → taint indirection introduces a timing dependency. Can we instead set an uninitialized:NoBind taint unconditionally at Server creation (like the K8s CCM pattern) so there's no race?
- I would use conditions for gate status reporting and bridge those to taints via a controller. Gives better observability out of the box.
- I would not add a Manageable state. Gating is orthogonal to lifecycle. Taints + conditions handle this without touching the state machine.
- Re-gating is not addressed. Should at least be called out in non-goals.

afritzler:
The readiness gates and conditions should be generic. Network topology is just one example. Can we model this similar to Pod readiness gates? The Server spec declares which conditions must be met, any controller can own and reconcile its respective condition, and the gating controller only checks if all are True before removing the taint. Something like:

Sample Server spec:
```yaml
spec:
  readinessGates:
    - conditionType: NetworkTopologyVerified
    - conditionType: FirmwareCompliant
    - conditionType: BIOSConfigured
  taints:
    - key: metal.ironcore.dev/uninitialized
      effect: NoBind
status:
  conditions:
    - type: NetworkTopologyVerified
      status: "False"
      reason: Pending
    - type: FirmwareCompliant
      status: "True"
      reason: VersionMatch
    - type: BIOSConfigured
      status: "False"
      reason: PendingSettings
```

That way you can add firmware validation, BIOS checks, or storage verification later without touching metal-operator or the gating controller logic. The gating controller is dumb. It just watches whether all readinessGates conditions are satisfied and removes the taint. Each gate controller independently owns its condition.

ieps/14-availability-gating.md
adracus: Some comments here:
* Agree, we should not have an additional state. Per definition, the server is available, but some checks have failed -> Tainted
* For the unconditional taint setting: I'd actually leave this out. The `node.cloudprovider.kubernetes.io/uninitialized:NoSchedule` taint is set by the cloud provider on a `Node` until it's 'ready' according to some cloud provider internal rules. In our case, there is no cloud provider and if the metal operator was to set it unconditionally on creation, who would remove it? That's something I'd actually leave to an extensible server readiness mechanism.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,164 @@
...
+
+External services register a gate by setting a label on the `BMC` object:
+
+```yaml
+metadata:
+  labels:
+    metal.ironcore.dev/gate-network: "true"
+```
+
+The label key format is `metal.ironcore.dev/gate-<topic>: "true"`. Multiple labels can be present simultaneously.
```
adracus: Boolean labels / annotations are not extensible / flexible enough (think of wanting to specify per gate whether to re-apply a taint when it fails) and require some manual parsing logic that may or may not fail.

Instead, I'd recommend taking a similar approach as the node readiness controller: https://kubernetes.io/blog/2026/02/03/introducing-node-readiness-controller/

Introduce a separate resource that specifies what condition has to be present on a Server and what taint to apply otherwise. Have a central controller monitor these conditions. If we implement it similar to the node readiness controller, we can also re-apply the taints if a server ever fails these checks at runtime, allowing us to immediately remediate having a faulty node.

adracus (changes requested):
I like the proposal of gating a server flexibly very much! We can push it to be more flexible and tool agnostic with some adaptations to match the style in k8s's node readiness controller ( see https://kubernetes.io/blog/2026/02/03/introducing-node-readiness-controller/ ).

While we know of the agora use case, I'd also recommend to not explicitly name 3rd party tools as required components and only describe what 3rd party tools are ought to do. We can of course still name 3rd party tools in the motivation / rationale.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,280 @@
+---
+title: Availability Gating for IronCore Servers
```
afritzler: I would suggest to use here: "Introducing ServerReadinessRule to ensure Server readiness"

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,280 @@
...
+
+Each controller has exclusive ownership of its taint key. They never interact with each other's taints.
+
+### Bootstrap taint
+
+Metal-operator gains a `--availability-gating` flag that accepts a comma-separated list of trigger points:
+
+```
+--availability-gating=creation,maintenance,release
+```
```
afritzler: `--readiness-gating=true` (simple toggle), or no flag at all. Gating is active whenever matching ServerReadinessRules exist.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,280 @@
...
+- Defining the logic run by individual gate controllers
+- Replacing existing maintenance workflows
+- Automatic timeout or force-removal of taints for stuck gates — operators should alert on long-gated servers via the provided metrics
+- Eviction of already-reserved servers based on gate state
+
+## Proposal
+
+### Overview
+
+Gating is enforced via `NoBind` taints on the `Server` object. A server can only be claimed when all `NoBind` taints are gone. There are two categories of taint:
```
afritzler: NoClaim directly maps to the action being prevented (a ServerClaim cannot claim this Server). NoBind is ambiguous in a domain with multiple binding concepts.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,280 @@
...
+spec:
+  serverSelector:
+    matchLabels:
+      metal.ironcore.dev/gate-network: "true"
+  condition:
+    type: NetworkTopologyVerified
+  taint:
+    key: metal.ironcore.dev/network-not-ready
+    effect: NoBind
+  triggers: [creation, maintenance]
```
afritzler: Removed entirely. If a rule matches and the condition is not met, the taint is applied regardless of lifecycle event.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,280 @@
...
+---
+apiVersion: metal.ironcore.dev/v1alpha1
+kind: ServerReadinessRule
+metadata:
+  name: disk-readiness
+spec:
+  serverSelector:
+    matchLabels:
+      metal.ironcore.dev/gate-disk: "true"
```
afritzler: Use existing labels (size, site, hardware type) in rule selectors.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,280 @@
...
+    type: DiskSanitized
+  taint:
+    key: metal.ironcore.dev/disk-not-ready
+    effect: NoBind
+  triggers: [release]
+```
+
+The `triggers` field declares which bootstrap taint values the gate controller responds to. When metal-operator sets the bootstrap taint, the gate controller checks whether the taint value matches its declared triggers before applying its gate taint and running its check.
+
+### ServerReadinessGate CRD
```
afritzler: Conditions already tell you which gates pass. Taints tell you if the server is claimable. A dedicated CRD for this is over-engineering and adds GC burden.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,280 @@
...
+```
+metaloperator_server_nobind_taint{server="compute-01", namespace="default", taint="metal.ironcore.dev/not-ready"} 1
+metaloperator_server_nobind_taint{server="compute-01", namespace="default", taint="metal.ironcore.dev/network-not-ready"} 1
+```
+
+Aggregate metric — count of servers per taint key:
+
+```
+metaloperator_server_nobind_taint_total{taint="metal.ironcore.dev/not-ready"} 5
+metaloperator_server_nobind_taint_total{taint="metal.ironcore.dev/network-not-ready"} 3
```
afritzler: Violates Prometheus naming conventions. Use metaloperator_server_noclaim_taint_count or rely on PromQL aggregation over the per-server metric.

afritzler (commented):
1. TOC typo [ServerReadinesssRule CRD](#serverreadinesssrule-crd) has a triple-s. Broken link.
2. Stale external reporters: If an external controller crashes, its condition stays False forever and the server is stuck indefinitely. No recommended pattern for detecting this (e.g., lastProbeTime on conditions, alerting on conditions stuck False beyond a threshold).
3. Rule deletion with orphaned taints: What happens when a ServerReadinessRule is deleted while servers still carry its taint? Does metal-operator garbage-collect the orphaned taint, or does the server stay gated on a rule that no longer exists?
4. Condition staleness: An external controller sets a condition to True, then crashes. The condition remains True indefinitely but may no longer reflect reality. No TTL or refresh mechanism is proposed.
5. Taint key conflicts: What if two ServerReadinessRules declare the same taint key with different conditions? Which one wins? Should this be validated (webhook rejecting duplicates)?
6. Condition type collisions: What if an external controller reports a condition type that matches one of metal-operator's known internal types (e.g., MemoryHealthVerified)? Who wins? Should internal types be reserved/prefixed?
7. Server enters Reserved while carrying taints. The original says "NoBind taints do not affect Reserved servers" but doesn't specify whether taints are cleaned up on reservation or left in place. Clarify the invariant: are taints preserved across the Reserved state for re-evaluation on release?
8. Rule selector changes: A rule's serverSelector is updated and a previously-matching server no longer matches. Does metal-operator remove the now-irrelevant taint?

ieps/14-availability-gating.md
xkonni: gating active when matching `ServerReadinessRules` exist sounds good to me. done.

ieps/14-availability-gating.md
xkonni: `NoClaim` it is.

ieps/14-availability-gating.md
xkonni: and leave the decision to the external controllers. alright!

ieps/14-availability-gating.md
xkonni: done.

ieps/14-availability-gating.md
xkonni: removed.

ieps/14-availability-gating.md
xkonni: fixed and going with PromQL.

xkonni:
> 1. TOC typo [ServerReadinesssRule CRD](#serverreadinesssrule-crd) has a triple-s. Broken link.
> 2. Stale external reporters: If an external controller crashes, its condition stays False forever and the server is stuck indefinitely. No recommended pattern for detecting this (e.g., lastProbeTime on conditions, alerting on conditions stuck False beyond a threshold).
> 3. Rule deletion with orphaned taints: What happens when a ServerReadinessRule is deleted while servers still carry its taint? Does metal-operator garbage-collect the orphaned taint, or does the server stay gated on a rule that no longer exists?
> 4. Condition staleness: An external controller sets a condition to True, then crashes. The condition remains True indefinitely but may no longer reflect reality. No TTL or refresh mechanism is proposed.
> 5. Taint key conflicts: What if two ServerReadinessRules declare the same taint key with different conditions? Which one wins? Should this be validated (webhook rejecting duplicates)?
> 6. Condition type collisions: What if an external controller reports a condition type that matches one of metal-operator's known internal types (e.g., MemoryHealthVerified)? Who wins? Should internal types be reserved/prefixed?
> 7. Server enters Reserved while carrying taints. The original says "NoBind taints do not affect Reserved servers" but doesn't specify whether taints are cleaned up on reservation or left in place. Clarify the invariant: are taints preserved across the Reserved state for re-evaluation on release?
> 8. Rule selector changes: A rule's serverSelector is updated and a previously-matching server no longer matches. Does metal-operator remove the now-irrelevant taint?

Thanks for the thorough review. Addressed all 8 points:

1. TOC typo — fixed.

2. Stale external reporters — added lastTransitionTime to all condition examples. The observability section now calls out alerting on conditions stuck False beyond a threshold as the recommended pattern. Automatic timeout/force-ungating stays a non-goal — a stale False correctly keeps the server gated.

3. Rule deletion / orphaned taints — metal-operator now adds a metal.ironcore.dev/serverreadinessgc finalizer to each ServerReadinessRule at creation. On deletion, it removes the rule's taint from all matching servers before releasing the finalizer. Gate controller availability does not block cleanup.

4. Condition staleness — added an optional condition.maxAge field to the rule for recurring gates. Metal-operator re-applies the gate taint when lastTransitionTime exceeds maxAge. Also added an explicit re-arm contract: gate controllers must reset their condition to False when re-arming.

5. Taint key conflicts — a validating admission webhook rejects any ServerReadinessRule whose spec.taint.key is already claimed by another rule.

6. Condition type collisions — the same webhook also rejects rules referencing internal metal-operator condition types, which are reserved under the metal.ironcore.dev/ prefix.

7. Reserved state / taints — clarified the invariant: taints added during Reserved (e.g. hardware degradation) are preserved and re-evaluated when the release trigger fires on server release.

8. Rule selector changes — handled by the same finalizer logic: on spec.serverSelector update, metal-operator removes the rule's taint from servers that no longer match the updated selector.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,271 @@
...
+metaloperator_server_noclaim_taint{server="compute-01", namespace="default", taint="metal.ironcore.dev/network-not-ready"} 1
+```
+
+Aggregate counts can be derived via PromQL:
+
+```promql
+count by (taint) (metaloperator_server_noclaim_taint)
+```
+
+Gate controller conditions on `Server.status.conditions` include `lastTransitionTime`. Operators can alert on conditions that have remained `False` beyond a deployment-specific threshold (e.g. 2 hours) using a `kube_customresource_status_condition` query. A condition stuck `False` means the gate controller has not cleared it — either the check is genuinely failing or the gate controller has crashed. In either case the server correctly remains gated; the alert surfaces which servers need attention.
```
afritzler: Do we really need the monitoring aspect in this proposal? I would add this to the docs once implemented.

afritzler (dismissed):
@xkonni thanks a lot for the update! Just one nit from my side.

ieps/14-availability-gating.md
xkonni: no, we can do it later. just thought we'd decide on it now.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,271 @@
...
+## Proposal
+
+### Overview
+
+Gating is enforced via `NoClaim` taints on the `Server` object. A server can only be claimed when all `NoClaim` taints are gone. There are two categories of taint:
+
+- **Bootstrap taint** (`metal.ironcore.dev/not-ready`) — set by metal-operator at trigger points, removed by metal-operator once no other `NoClaim` taints remain
+- **Gate taints** (e.g. `metal.ironcore.dev/network-not-ready`) — set and removed by external gate controllers, one taint key per gate type
+
+Each controller has exclusive ownership of its taint key. They never interact with each other's taints.
```
adracus: This section is correct in what the readiness rules can be used for, but since the mechanism is now more generic, we should phrase this section in a more generic way as well.

Additionally, the API itself does not impose any guarantees on who sets which taint - it's perfectly
possible that anybody with permissions can create a rule that e.g. applies the `metal.ironcore.dev/not-ready` taint without them being the `metal-operator`. Thus, I'd also recommend removing the section
about who 'owns' which taints completely.

Also, a server could potentially be claimed even if it has taints on it if the claim has the sufficient tolerations - we should factor that in as well.

How about stating something like

```suggestion
The taints of `Server` objects can be manipulated via `ServerReadinessRule` objects. These rules
monitor the `.status.conditions` of a `Server` for a specific condition type and status.

If the specific condition type is not present in the expected status, a taint specified in the rule is
applied to the `Server`.

Once the condition is present in the expected status, the taint is removed.

This makes a `Server` unclaimable to `ServerClaim`s that don't have the required tolerations.
```

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,271 @@
...
+    - [ServerReadinessRule CRD](#serverreadinessrule-crd)
+    - [Lifecycle example](#lifecycle-example)
+    - [Re-gating](#re-gating)
+    - [Observability](#observability)
+- [Dependencies](#dependencies)
+- [Related](#related)
+
+## Summary
+
+Introduce a generic availability gating mechanism that prevents servers from being claimed until all declared readiness gates pass. The design follows the upstream [Kubernetes Node Readiness Controller](https://kubernetes.io/blog/2026/02/03/introducing-node-readiness-controller/) pattern: metal-operator sets a bootstrap taint at server creation, external gate controllers independently manage their own gate-specific taints and report conditions on the `Server` object, and metal-operator removes the bootstrap taint once all gate taints are cleared.
```
adracus: `metal-operator` should not set an own specific bootstrap taint at server creation. It should still 'just' do its default inspection to say when a server is ready and use it for scheduling once it's ready. The rules however allow to set a taint depending on specific conditions and also remove them automatically once the conditions are present in the expected status. This also means that there are no external controllers touching the taints of a server but only writing specific conditions on which the rules react.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,271 @@
...
+## Motivation
+
+Servers entering the fleet may fail pre-availability checks (e.g. physical wiring, unclean disks). Without a gating mechanism, a server can reach `Available` state and get claimed before these checks pass, leading to issues at workload runtime.
+
+The driving use case is network topology verification: an external controller compares discovered interface data against an authoritative source and only clears its gate condition once the topology matches. Other gate types follow the same pattern without requiring changes to metal-operator.
+
+### Goals
+
+- Block servers from being claimed until all declared readiness gates pass
+- Support multiple concurrent gate types owned by independent controllers
```
adracus, line 47: I'd remove the part about controllers - it's about 'entities' in general.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,271 @@
...
+
+Metal-operator removes the bootstrap taint once no other `NoClaim` taints remain on the server. The expected flow is that external gate controllers apply their gate taints early in Discovery (which takes several minutes), so by the time Discovery completes the gate taints are in place. If no gate controllers target the server, the bootstrap taint is removed at the end of Discovery and the server proceeds normally.
+
+**`NoClaim` taints do not affect `Reserved` servers.** The taint is only evaluated at claim time — it prevents a new `ServerClaim` from binding but does not evict an existing reservation. Taints added while a server is `Reserved` (e.g. hardware degradation detected during a workload's lifetime) are preserved and not cleaned up on reservation. When the server is released and transitions back to `Available`, the `release` trigger fires: metal-operator re-sets the bootstrap taint and all matching gate controllers re-evaluate. The server becomes claimable again only after all taints are cleared.
```
adracus: We should adapt this more to how the node readiness controller handles the enforcement modes: https://node-readiness-controller.sigs.k8s.io/user-guide/concepts.html?highlight=enforcement-md - this would also require to describe the enforcement mode + implications in the proposal.

The metal operator sets taints for the rules that are required during bootstrap and also applies the correct annotations once the rule has been satisfied. We shouldn't encode special meaning into the values of the taints.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,271 @@
...
+Gate controllers do not interact with the bootstrap taint or with each other.
+
+**Re-arm contract:** when a gate controller re-arms at a trigger point (i.e. it applies its gate taint because the bootstrap taint appeared on a matching server), it must reset its condition to `False` before running its check. This ensures a passing condition from a previous cycle does not carry over and cause the gate to be skipped.
+
+Servers are brought into scope via the `spec.serverSelector` in the `ServerReadinessRule` — this is the gate controller's targeting configuration. Any labels already present on the server (hardware type, region, zone, etc.) can be used; no dedicated gate labels are required.
+
+### ServerReadinessRule CRD
+
+A `ServerReadinessRule` is a cluster-scoped resource, created once by whoever deploys a gate controller. It declares which servers the gate controller targets, which condition it monitors, and which taint it manages. Gate controllers read rules to determine which servers to target and which taint to manage. Metal-operator watches rules solely for lifecycle management: it adds a finalizer at creation and removes orphaned gate taints on deletion or selector change.
```
adracus: A server readiness rule does not have any dependency on a gate controller. It's standalone and can be created independently. A controller writing a condition on a `Server` might not even know that there's a readiness rule imposed on a condition.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,271 @@
...
+    matchLabels:
+      example.com/hardware-type: server-gen2
+  condition:
+    type: DiskSanitized
+  taint:
+    key: metal.ironcore.dev/disk-not-ready
+    effect: NoClaim
+```
+
+The optional `condition.maxAge` field marks a gate as recurring. Metal-operator monitors `lastTransitionTime` on the corresponding server condition and re-applies the gate taint if the condition has been `True` for longer than `maxAge`, signalling the gate controller to re-run. Gate controllers for recurring gates are expected to periodically re-evaluate and re-patch their condition to keep `lastTransitionTime` current. If `maxAge` is absent the gate is treated as one-shot: the condition persists until the gate controller explicitly changes it or a trigger point resets it.
```
adracus: We should have the same enforcement modes as the node readiness controller (bootstrap only / continuous) here.

adracus (changes requested):

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,271 @@
...
+- Automatic timeout or force-removal of taints for stuck gates — operators should alert on long-gated servers via the provided metrics
+- Eviction of already-reserved servers based on gate state
+
+## Proposal
+
+### Overview
+
+Gating is enforced via `NoClaim` taints on the `Server` object. A server can only be claimed when all `NoClaim` taints are gone. There are two categories of taint:
+
+- **Bootstrap taint** (`metal.ironcore.dev/not-ready`) — set by metal-operator at trigger points, removed by metal-operator once no other `NoClaim` taints remain
```
stefanhipfel: could there be a scenario when a server's discovery is too quick, before another entity has time to add its taints?
So the servers becomes Initial->Available and after Available an entity finally adds its taints.
Sure the Server is still not claimable, but I think the status Available in this case could be confusing?!

xkonni: i do not think so. as discussed in the meeting yesterday, once the `serverreadinessrules` exist, the taints will be re-added by metal-operator when passing the trigger-points.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,310 @@
...
+    - metal.ironcore.dev/serverreadinessgc
+spec:
+  serverSelector:
+    matchLabels:
+      example.com/hardware-type: server-gen2
+  enforcement:
+    mode: bootstrap
+    triggers:
+      - creation
+      - maintenance
```
adracus: We should just have a plain value like

```yaml
enforcementMode: BootstrapOnly
```

or

```yaml
enforcementMode: Continous
```

and drop `triggers`. Each rule is evaluated at each change of the server.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,310 @@
...
+  condition:
+    type: DiskSanitized
```
adracus: As in the node readiness proposal, make this `conditions` + also require the status the condition must have to satisfy the rule like


```suggestion
  conditions:
  - type: DiskSanitized
    requiredStatus: True
  - type: DiskPressure
    requiredStatus: False
```

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,310 @@
...
+  condition:
+    type: HardwareInventoryVerified
+  taint:
+    key: metal.ironcore.dev/hardware-not-verified
+    effect: NoClaim
+```
+
+**Taint key uniqueness** is enforced by a validating admission webhook. On create and update of a `ServerReadinessRule`, the webhook rejects the request if any other rule declares the same `spec.taint.key`. Each taint key must be owned by exactly one rule.
+
+**Reserved condition types** are enforced by the same webhook. `ServerReadinessRule` objects whose `spec.condition.type` uses the `metal.ironcore.dev/` prefix are rejected — that namespace is reserved for metal-operator internal conditions.
```
adracus: Why this? I personally don't see a reason to have reserved metal-operator internal conditions.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,310 @@
...
+
+- At creation, adds `metal.ironcore.dev/serverreadinessgc` as a finalizer.
+- On deletion, removes the rule's taint from all servers that carry it, then releases the finalizer.
+- On `spec.serverSelector` change, removes the rule's taint from servers that no longer match the updated selector.
+
+### Enforcement modes
+
+#### `bootstrap`
+
+The rule is enforced at specific trigger points only. At each listed trigger, the `ReadinessGateController` applies the taint to all matching servers. Once the condition passes and the taint is removed, the controller records completion via annotation and stops enforcing that rule for that server until the next trigger fires.
```
adracus: We don't specify trigger points. Bootstrap only means it should only monitor that condition until it's reached and (as you wrote correctly) write an annotation that marks it as 'completed'.

```suggestion
Specifically for one-time initialization steps, such as validating configuration or hardware provisioning. Once conditions are met, the controller marks the bootstrap as complete and stops monitoring that specific rule for the server.
```

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,310 @@
...
+```
+
+If the condition later flips to `False` at runtime, nothing happens — the taint is not re-applied until the next trigger point.
+
+Supported triggers:
+
+| Trigger | When it fires |
+|---|---|
+| `creation` | Server enters Discovery for the first time |
+| `maintenance` | Server exits Maintenance state |
```
adracus: Remove triggers. If `enforcementMode: BootstrapOnly` the server will be evaluated each time it changes and once it satisfies the condition, it won't be anymore. If it's `enforcementMode: Continuous` it will monitor it continuously and at every change evaluate the rule, adding / removing the taint every time if necessary.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,310 @@
...
+### ServerReadinessRule CRD
+
+A `ServerReadinessRule` is a cluster-scoped resource that declares:
+
+- Which servers to target (`spec.serverSelector`)
+- Which condition to monitor (`spec.condition.type`)
+- Which taint to manage (`spec.taint`)
+- When to enforce the rule (`spec.enforcement`)
+
+```yaml
```
adracus, line 91: Please let the reader know these are example manifests and not part of the metal operator.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,310 @@
...
+
+#### `continuous`
+
+The rule is always enforced regardless of trigger. If the condition is not satisfied at any point, the `ReadinessGateController` applies the taint immediately. When the condition recovers, the taint is removed. Suitable for checks that must hold throughout the server's lifetime.
+
+### Lifecycle examples
+
+#### Server created
+
+All three rules apply at `creation`. `ReadinessGateController` applies all three taints immediately.
```
adracus:
```suggestion
Looking at the sample rules above, all three rules match the `Server` and are applied upon creation of the `Server`. The `ReadinessGateController` applies all three taints immediately.
```

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,310 @@
...
+      effect: NoClaim
+    - key: metal.ironcore.dev/network-not-ready       # applied by ReadinessGateController
+      effect: NoClaim
+    - key: metal.ironcore.dev/hardware-not-verified   # applied by ReadinessGateController
+      effect: NoClaim
+status:
+  conditions: []
+```
+
+Hardware inventory controller runs first and sets its condition `True`. `ReadinessGateController` removes the hardware taint and annotates bootstrap completion.
```
adracus: Please remove any mention of specific external controllers. The proposal should be about `metal-operator` only.

```suggestion
An external controller sets the condition `HardwareInventoryVerified` condition to `True`. `ReadinessGateController` removes the hardware taint and annotates bootstrap completion.
```

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,310 @@
...
+      effect: NoClaim
+status:
+  conditions:
+    - type: HardwareInventoryVerified
+      status: "True"
+      reason: Verified
+      lastTransitionTime: "2026-05-29T10:03:00Z"
+```
+
+Disk sanitization controller sets its condition `True`. `ReadinessGateController` removes the disk taint.
```
adracus: Same as above.

```suggestion
Another external controller sets the `DiskSanitized` condition to `True`. `ReadinessGateController` removes the disk taint.
```

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,310 @@
...
+      status: "True"
+      reason: Verified
+      lastTransitionTime: "2026-05-29T10:03:00Z"
+    - type: DiskSanitized
+      status: "True"
+      reason: Sanitized
+      lastTransitionTime: "2026-05-29T10:05:00Z"
+```
+
+Network topology controller sets its condition `True`. `ReadinessGateController` removes the last taint. Server is claimable.
```
adracus:
```suggestion
Finally, another controller sets the `NetworkTopologyVerified` condition to `True`. `ReadinessGateController` removes the last taint. Server is claimable.
```

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,310 @@
...
+
+### Observability
+
+Metal-operator exposes a per-server metric — one time series per server/taint combination, value `1` while active:
+
+```
+metaloperator_server_noclaim_taint{server="compute-01", taint="metal.ironcore.dev/disk-not-ready"} 1
+```
+
+Additional metrics may be added as the implementation matures.
```
adracus: I'd remove this section for now. If we want to introduce metrics, I'd also base our approach more on how the node readiness controller does it: https://node-readiness-controller.sigs.k8s.io/operations/monitoring.html

adracus (changes requested):

adracus:
@xkonni could you please take a look at the open comments?

ieps/14-availability-gating.md
xkonni: alright. thought as a safeguard that would be a good thing. removing that section.

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,245 @@
...
+  finalizers:
+    - metal.ironcore.dev/serverreadinessgc
+spec:
+  serverSelector:
+    matchLabels:
+      example.com/hardware-type: server-gen2
+  enforcement:
+    mode: bootstrap
+  condition:
+    type: ExampleCondition
```
adracus:
```suggestion
  conditions:
  - type: ExampleCondition
    status: True
```

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,245 @@
...
+  finalizers:
+    - metal.ironcore.dev/serverreadinessgc
+spec:
+  serverSelector:
+    matchLabels:
+      example.com/hardware-type: server-gen2
+  enforcement:
+    mode: continuous
+  condition:
+    type: ExampleContinuousCondition
```
adracus:
```suggestion
  conditions:
  - type: ExampleContinuousCondition
    status: True
```

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,245 @@
...
+- Manages rule lifecycle: adds a finalizer at rule creation, removes orphaned taints on rule deletion or selector change
+
+Any entity with the appropriate permissions can write conditions on `Server.status.conditions`. They do not need to apply or remove taints and need not be aware of any `ServerReadinessRule` that references their condition type.
+
+### ServerReadinessRule CRD
+
+A `ServerReadinessRule` is a cluster-scoped resource that declares:
+
+- Which servers to target (`spec.serverSelector`)
+- Which condition to monitor (`spec.condition.type`)
```
adracus:
```suggestion
- Which condition to monitor (`spec.conditions[*].type` + `spec.conditions[*].status`)
```

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,245 @@
...
+- Which taint to manage (`spec.taint`)
+- How to enforce the rule (`spec.enforcement.mode`)
```
adracus:
```suggestion
- How to enforce the rule (`spec.enforcementMode`)
```

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,245 @@
...
+metadata:
+  name: example-one-shot
+  finalizers:
+    - metal.ironcore.dev/serverreadinessgc
+spec:
+  serverSelector:
+    matchLabels:
+      example.com/hardware-type: server-gen2
+  enforcement:
+    mode: bootstrap
```
adracus:
```suggestion
  enforcementMode: BootstrapOnly
```

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,245 @@
...
+metadata:
+  name: example-continuous
+  finalizers:
+    - metal.ironcore.dev/serverreadinessgc
+spec:
+  serverSelector:
+    matchLabels:
+      example.com/hardware-type: server-gen2
+  enforcement:
+    mode: continuous
```
adracus:
```suggestion
  enforcementMode: Continuous
```

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,245 @@
...
+
+**Rule lifecycle managed by ReadinessGateController:**
+
+- At creation, adds `metal.ironcore.dev/serverreadinessgc` as a finalizer.
+- On deletion, removes the rule's taint from all servers that carry it, then releases the finalizer.
+- On `spec.serverSelector` change, removes the rule's taint from servers that no longer match the updated selector.
+
+### Enforcement modes
+
+#### `bootstrap`
```
adracus:
```suggestion
#### `BootstrapOnly`
```

ieps/14-availability-gating.md
```diff
@@ -0,0 +1,245 @@
...
+
+```yaml
+metadata:
+  annotations:
+    metal.ironcore.dev/bootstrap-completed-example-one-shot: 'true'
+```
+
+If the condition later flips to `False` at runtime, nothing happens — the taint is not re-applied. Re-gating can be triggered externally by clearing the completion annotation, at which point the controller picks up the server again on its next reconcile.
+
+#### `continuous`
```
adracus:
```suggestion
#### `Continuous`
```

adracus (changes requested):

adracus:
@xkonni somehow the enhancement for specifying multiple conditions slipped through + the rename of `enforcement: mode` to `enforcementMode` and the values should be (according to k8s api conventions) `BootstrapOnly` + `Continuous` - could you please adapt so we can get this in?

adracus (approved):
Thanks for the nice proposal 🚀

PR #21, opened 2026-06-01
Wire proposal

adracus:
This proposes the wire API for configuring switches.

### DIFF
--- a/ieps/tbd-wire.md
+++ b/ieps/tbd-wire.md
@@ -1,0 +1,465 @@
+---
+title: Wire
+
+iep-number: tbd
+
+creation-date: 2026-05-29
+
+status: review
+
+authors:
+
+- "@adracus"
+- "@afritzler"
+
+reviewers:
+  
+- tbd
+
+---
+
+# IEP-tbd: Wire
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Alternatives](#alternatives)
+
+## Summary
+
+Implement an extensible and declarative switch network API that expresses
+how we configure our network.
+
+## Motivation
+
+We currently provision or switches with central templates and lots
+of implicit configuration. Instead of this, we want to make the configuration
+more explicit while staying vendor independent. Additionally, we want to
+have a clear language on how we configure our switches and when.
+
+Frequent reconfiguration of switches during runtime has shown in countless
+examples that it jeopardizes the stability of the network. As such,
+reconfiguration must be brought to a minimum and made absolutely explicit
+when it happens.
+
+### Goals
+
+* Stay vendor-independent
+* Declarative, making it absolutely explicit and significant when switches
+  are (re)configured.
+
+### Non-Goals
+
+* Non-stable state for our switches - A configuration must be applied
+  once and not be continuously re-evaluated, causing unexpected side effects.
+* Imperative API - The network should be declarative.
+
+## Proposal
+
+### Current State
+
+Our network follows a CLOS topology: A level of spines, leaves
+and hosts connected to the leaves.
+
+Each host is connected to two leaves and each leave is connected to
+two spines for extra redundancy.
+
+The connection towards the hosts from the leaves is also 'wrapped'
+with a VLAN per interface since only by using VLANs, DHCP relay
+can be specified. This is necessary to be able to boot servers
+using network boot (PXE / HTTP).
+
+Each member of the topology (spine, leaf, host) runs BGP unnumbered
+for route distribution.
+
+Diagram of an excerpt of how our network looks like:
+
+```mermaid
+graph TD
+    spine-01["`**spine-01**
+    /64 prefix
+    /128 loopback`"]
+    spine-01-if-01
+    spine-01-if-02
+
+    leaf-01["`**leaf-01**
+    /64 prefix
+    /128 loopback`"]
+    leaf-01-vlan-host-01["`VLAN /80`"]
+
+    leaf-02["`**leaf-02**
+    /64 prefix
+    /128 loopback`"]
+    leaf-02-vlan-host-01["`VLAN /80`"]
+
+    host-01["`**host-01**
+    /64 prefix
+    /128 loopback`"]
+
+    dhcp["DHCP"]
+
+    %% Spines
+    subgraph Spine 01
+    spine-01---spine-01-if-01
+    spine-01---spine-01-if-02
+    end
+
+    %% Leafs
+    subgraph Leaf 01
+    leaf-01-if-01---leaf-01
+    leaf-01---leaf-01-if-02
+    leaf-01---leaf-01-if-03
+    leaf-01-if-02---leaf-01-vlan-host-01
+    end
+
+    subgraph Leaf 02
+    leaf-02-if-01---leaf-02
+    leaf-02---leaf-02-if-02
+    leaf-02---leaf-02-if-03
+    leaf-02-if-02---leaf-02-vlan-host-01
+    end
+
+    %% Hosts
+    subgraph Host 01
+    host-01-if-01---host-01
+    host-01-if-02---host-01
+    end
+
+    %% VLANs to DHCP
+    leaf-01-vlan-host-01---dhcp
+    leaf-02-vlan-host-01---dhcp
+
+    %% Spines to Leafs
+    spine-01-if-01 <-->leaf-01-if-01
+
+    spine-01-if-02 <-->leaf-02-if-01
+
+
+    %% Leafs to Hosts
+    leaf-01-vlan-host-01 <-->host-01-if-01
+
+    leaf-02-vlan-host-01<-->host-01-if-02
+```
+
+For configuring our switches we currently use ZTP (zero-touch-provisioning).
+We render the configuration from a template, since we know how our
+cabling looks like. To keep the template small we use BGP unnumbered,
+allowing us to omit each neighbor's ASN number.
+
+This setup configures switches *once*, avoiding frequent switch
+reconfiguration. Switch reconfiguration is known to be one of the
+core issues causing severe network disruption.
+
+### Declarative Network Design
+
+When designing the resources for a new API, we start with
+the ground truths we know about: Our cabling / topology plan.
+
+At the cluster scope, we define the following types:
+
+* **`Device`** representing an 'unconfigured' switch.
+* **`DeviceInterface`** representing an interface of a device.
+* **`Server`** representing an arbitrary host.
+* **`ServerInterface`** representing an interface of a server.
+
+We distinguish between devices and servers since devices and their
+interfaces can be reconciled (e.g. set admin state `Up` / `Down`,
+apply some configuration) while servers are externally managed (e.g.
+by their OS / config).
+
+The cluster-scoped **`Link`** resource specifies which interface
+(either `DeviceInterface` or `ServerInterface`) is linked to each other.
+For each such connection, a `Link` is created.
+
+To actually make a `Device` function as a switch inside the network,
+routing traffic properly, a namespaced `Switch` resource is created.
+This `Switch` resource references a `Device` and, once accepted by
+the `Device`, the `Device` references the `Switch` back. The spec
+of a `Switch` is immutable. Once a `Switch` is created, this causes
+the underlying device to be reconfigured. By having a dedicated `Switch`
+resource, we gain several core benefits:
+
+* Reconfiguration becomes explicit: We do not want to continuously
+  reconfigure a switch but only configure it as seldom as possible.
+  A single object contains everything needed to configure the switch.
+
+* Having a single object means the implementors of this API can construct
+  the most optimal way to apply the entire configuration: E.g. depending
+  on the vendor, the sequence to apply a configuration can heavily differ.
+  Having the entire desired configuration at once is the only allow a vendor
+  to implement this correctly.
+
+* By having a `Switch` resource that expresses the effective configuration,
+  rolling / draining traffic and gracefully switching between two `Switch`
+  configurations can be done. One could e.g. think of a higher-level type
+  and controller that first drains traffic, removes the old `Switch` object
+  once drained and creates a new one once ready.
+
+Since servers also cooperatively take part in our networking, a `Host` resource
+is what a `Switch` is to a `Device`: It expresses how a `Host` should join the network.
+In contrast to a `Switch` however, this is not reconciled on a switch level but
+depends on the server itself.
+
+### Sample Resources
+
+Spine (cluster-scoped):
+
+```yaml
+apiVersion: wire.ironcore.dev
+kind: Device
+metadata:
+    name: spine-01
+spec:
+  providerID: sonic://spine-01
+---
+apiVersion: wire.ironcore.dev
+kind: DeviceInterface
+metadata:
+  name: spine-01-if-01
+spec:
+  handle: sonic://if-01
+  adminState: Up
+  deviceRef:
+    name: spine-01
+```
+
+Leaf (cluster-scoped):
+
+```yaml
+apiVersion: wire.ironcore.dev
+kind: Device
+metadata:
+    name: leaf-01
+spec:
+  providerID: sonic://leaf-01
+---
+apiVersion: wire.ironcore.dev
+kind: DeviceInterface
+metadata:
+  name: leaf-01-if-01
+spec:
+  handle: sonic://if-01
+  adminState: Up
+  deviceRef:
+    name: leaf-01
+---
+apiVersion: wire.ironcore.dev
+kind: DeviceInterface
+metadata:
+  name: leaf-01-if-02
+spec:
+  handle: sonic://if-02
+  adminState: Up
+  deviceRef:
+    name: leaf-01
+```
+
+Server (cluster-scoped):
+
+```yaml
+apiVersion: wire.ironcore.dev
+kind: Server
+metadata:
+    name: host-01
+---
+apiVersion: wire.ironcore.dev
+kind: ServerInterface
+metadata:
+  name: host-01-if-01
+serverRef:
+  name: host-01
+```
+
+Links (cluster-scoped):
+
+```yaml
+apiVersion: wire.ironcore.dev
+kind: Link
+metadata:
+    name: spine-01-if-01-leaf-01-if-01
+endpoints:
+- deviceInterfaceRef:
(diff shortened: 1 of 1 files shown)


ieps/tbd-wire.md
```diff
@@ -0,0 +1,465 @@
...
+- [Alternatives](#alternatives)
+
+## Summary
+
+Implement an extensible and declarative switch network API that expresses
+how we configure our network.
+
+## Motivation
+
+We currently provision or switches with central templates and lots
```
hardikdr, line 39:
```suggestion
We currently provision our switches with central templates and lots
```

ieps/tbd-wire.md
```diff
@@ -0,0 +1,465 @@
...
+
+### Non-Goals
+
+* Non-stable state for our switches - A configuration must be applied
+  once and not be continuously re-evaluated, causing unexpected side effects.
+* Imperative API - The network should be declarative.
+
+## Proposal
+
+### Current State
```
hardikdr, line 63: Curious whether anchoring the motivation to the current ironcore-underlay topology is intentional scoping, or whether this is meant to be a more general-purpose network automation API that happens to solve the ironcore use case first. 

If the latter, it might be worth framing the current state as an `initial use case` rather than the baseline,  so future use cases feel like natural extensions rather than deviations from the original design.

ieps/tbd-wire.md
```diff
@@ -0,0 +1,465 @@
...
+### Declarative Network Design
+
+When designing the resources for a new API, we start with
+the ground truths we know about: Our cabling / topology plan.
+
+At the cluster scope, we define the following types:
+
+* **`Device`** representing an 'unconfigured' switch.
+* **`DeviceInterface`** representing an interface of a device.
+* **`Server`** representing an arbitrary host.
```
hardikdr, line 166: IMO we dont need `Server` and `ServerInterface` as a dedicated custom-resource within network-automation, unless we have a `Spec` in mind that a controller can reconcile. While `Server` can be very useful for visualisers, or potential validation of the underlay, it might as well be a static input rather than a dedicated CRD.

ieps/tbd-wire.md
```diff
@@ -0,0 +1,465 @@
...
+* **`ServerInterface`** representing an interface of a server.
+
+We distinguish between devices and servers since devices and their
+interfaces can be reconciled (e.g. set admin state `Up` / `Down`,
+apply some configuration) while servers are externally managed (e.g.
+by their OS / config).
+
+The cluster-scoped **`Link`** resource specifies which interface
```
hardikdr, line 174: The same argument against `Server` goes for `Link` as well. I would recommend avoiding it unless we have a further `Spec` in mind to reconcile against `Link`. 

In `sonic-operator,` we used LLDP-based discovery, which helped identify the neighbours serving the same purpose inside the `Status` of `SwitchInterface`s, which could then be used for validation against desired topology, or visualisers.

ieps/tbd-wire.md
```diff
@@ -0,0 +1,465 @@
...
+
+To actually make a `Device` function as a switch inside the network,
+routing traffic properly, a namespaced `Switch` resource is created.
+This `Switch` resource references a `Device` and, once accepted by
+the `Device`, the `Device` references the `Switch` back. The spec
+of a `Switch` is immutable. Once a `Switch` is created, this causes
+the underlying device to be reconfigured. By having a dedicated `Switch`
+resource, we gain several core benefits:
+
+* Reconfiguration becomes explicit: We do not want to continuously
```
hardikdr, line 186: While separating the `Device` from `Switch` is a reasonable idea(thanks a lot for that), I find the complete immutability of the `Switch` inspired more by a very specific use case. Complete swap of the `Switch` for every change would force users to either not make any incremental changes or accept the potential network disruption with every change, which would be nice to avoid. 

In case there is an assumption that every Server is connected with 2 Leaves for redudancy, that is too topology-specific. Even in that case, a naturally degraded Switch would prevent the rollout of the other switch, as it could potentially affect a subset of the Servers.

ieps/tbd-wire.md
```diff
@@ -0,0 +1,465 @@
...
+  reconfigure a switch but only configure it as seldom as possible.
+  A single object contains everything needed to configure the switch.
+
+* Having a single object means the implementors of this API can construct
```
hardikdr, line 190: I would vouch for a more hybrid setup, where certain features, such as NTP, DNS, DHCP Relay, Loopback IPs, etc., could remain inline while operational features that can potentially have an independent lifecycle, have dedicated CRDs, such as BGP, etc.

ieps/tbd-wire.md
```diff
@@ -0,0 +1,465 @@
...
+  to implement this correctly.
+
+* By having a `Switch` resource that expresses the effective configuration,
+  rolling / draining traffic and gracefully switching between two `Switch`
+  configurations can be done. One could e.g. think of a higher-level type
+  and controller that first drains traffic, removes the old `Switch` object
+  once drained and creates a new one once ready.
+
+Since servers also cooperatively take part in our networking, a `Host` resource
+is what a `Switch` is to a `Device`: It expresses how a `Host` should join the network.
```
hardikdr, line 203: I would be critical about a dedicated `Host` CRD here, this would mean the network-automation to actively reconcile the `Host` and achieve the desired state. This leads network-automation to cross the structural boundary, I would keep the configuration of the Servers/Hosts separate and have them managed by the MetalAPI. 

Instead, some higher-level controller could later prepare the necessary network-related ignition-blob for Servers which then gets applied to the Servers via MetalAPI instead of network-automation handling it.

hardikdr (commented):
Thanks a lot for the proposal @adracus 
I have left a few comments inline.

hardikdr:
While reviewing this proposal, I ended up drafting an alternative design that builds on some ideas here - particularly the Device/Switch separation - while taking a different stance on immutability, object scope, and extensibility. Would be great to get everyone's thoughts on it as well: https://gist.github.com/hardikdr/5def28b4cb645d374d03f3d7fe65ab35

ieps/tbd-wire.md
```diff
@@ -0,0 +1,465 @@
...
+routing traffic properly, a namespaced `Switch` resource is created.
+This `Switch` resource references a `Device` and, once accepted by
+the `Device`, the `Device` references the `Switch` back. The spec
+of a `Switch` is immutable. Once a `Switch` is created, this causes
+the underlying device to be reconfigured. By having a dedicated `Switch`
+resource, we gain several core benefits:
+
+* Reconfiguration becomes explicit: We do not want to continuously
+  reconfigure a switch but only configure it as seldom as possible.
+  A single object contains everything needed to configure the switch.
```
peanball, line 188: to add to @hardikdr's point, I think reducing reconfiguration is a process goal, not something to be enforced technically. The single object containing the desired state does not contradict that. Different `Version`s of the same resource are equally valid.

ieps/tbd-wire.md
```diff
@@ -0,0 +1,465 @@
...
+
+We currently provision or switches with central templates and lots
+of implicit configuration. Instead of this, we want to make the configuration
+more explicit while staying vendor independent. Additionally, we want to
+have a clear language on how we configure our switches and when.
+
+Frequent reconfiguration of switches during runtime has shown in countless
+examples that it jeopardizes the stability of the network. As such,
+reconfiguration must be brought to a minimum and made absolutely explicit
+when it happens.
```
peanball, line 47: If this is a proposal for the whole WireAPI, i.e. the "network hardware bootstrapping and configurability", this shouldn't be a delta to whatever we are currently doing, but a desire for configuring a network that is useful for IronCore in general, and specific setups such as ours in particular.

The desire is to modify the configuration of switches as little as possible as they are pretty much "dumb but fast" network elements. Pretty much everything can talk to everything via the switches, i.e. we don't do explicit VLAN segmentation based on user workloads but only for setting the initial topology.

Any segmentation happens on the higher level via SDN, be it dpservice, an IPv6 only CNI or any other means. The WireAPI defines the fundamental network, on top of which other layers build.

The main goal of the WireAPI is then to express the desired topology and provide enough "knobs" to bootstrap the network from pretty much raw uninitialized hardware with minimal manual effort. There is still _some_ manual involvement necessary to get the raw hardware to the point that the WireAPI can talk to it.

There is no desire to continuously "converge" on a dynamically computed state (scheduling, etc.), but there is the desire for zero-downtime maintenance, which means that switches need to be taken out of rotation without interruption. There _may_ be a degradation in high availability during such maintenance activities, when one of the n (for us usually n=2) links goes down for a brief period while one of the switches is maintained.

The topology must be kept in mind for such maintenance activities, as we don't want to turn off all redundant connections to a specific target.

I fully agree with the last paragraph, as is. How we get there (described later in the proposal) is up for debate ;)

ieps/tbd-wire.md
```diff
@@ -0,0 +1,465 @@
...
+
+### Goals
+
+* Stay vendor-independent
+* Declarative, making it absolutely explicit and significant when switches
+  are (re)configured.
```
peanball, line 53: Maybe we can highlight that we don't want to have a full general-purpose-every-knob-exposed-via-gNMI-or-SNMP-configurability, but automation of a stable foundation network topology that is mostly stable with the exception of maintenance.

Automation focusses on initial setup and configuration (bootstrapping) to ensure that the desired topology is and remains in place.

ieps/tbd-wire.md
```diff
@@ -0,0 +1,465 @@
...
+
+### Non-Goals
+
+* Non-stable state for our switches - A configuration must be applied
+  once and not be continuously re-evaluated, causing unexpected side effects.
+* Imperative API - The network should be declarative.
```
peanball, line 59: This point is a bit confusing or rather too brief.

Imperative relates to _when_ something is set up, but not individual settings. We don't issue a command to set up some VLAN.

Declarative describes _what_ to set holistically, as a big picture. You can see the complete desired state at once.

Together, they mean that we set what we'd like the resource to look like. The imperative leans on the immutable resource, which is explicitly (re)created and then triggers the (re)configuration.

Maybe to consider whether we want a clear overview and traceability of what the system should look like and how it looks now, vs. manual operator control of _when_ things are set.

ieps/tbd-wire.md
```diff
@@ -0,0 +1,465 @@
...
+* **`ServerInterface`** representing an interface of a server.
+
+We distinguish between devices and servers since devices and their
+interfaces can be reconciled (e.g. set admin state `Up` / `Down`,
+apply some configuration) while servers are externally managed (e.g.
+by their OS / config).
+
+The cluster-scoped **`Link`** resource specifies which interface
+(either `DeviceInterface` or `ServerInterface`) is linked to each other.
+For each such connection, a `Link` is created.
```
peanball, line 176: I agree that this is more a "state" than a CR. The link is a physical property of someone plugging a cable in a data center. We can't will a link into existence by creating a CR, only discover/document the fact that the link exists.

ieps/tbd-wire.md
peanball: I would argue that some things such as BGP, addressing, etc. could actually be expressed as a "formula" and derived based on the desired topology.

If we know we want a Clos topology (as an example), then we know how we'd want hops to be in a systematic way. Then there is no need to explicitly, manually, configuring them.

ieps/tbd-wire.md
```diff
@@ -0,0 +1,465 @@
...
+  reconfigure a switch but only configure it as seldom as possible.
+  A single object contains everything needed to configure the switch.
+
+* Having a single object means the implementors of this API can construct
+  the most optimal way to apply the entire configuration: E.g. depending
+  on the vendor, the sequence to apply a configuration can heavily differ.
+  Having the entire desired configuration at once is the only allow a vendor
+  to implement this correctly.
+
+* By having a `Switch` resource that expresses the effective configuration,
```
peanball, line 196: This could equally be done with taints and resource versioning. It is still logically the same switch in the same place with the same name, but it is e.g. running an OS update. When it comes back it's still the same switch.

ieps/tbd-wire.md
```diff
@@ -0,0 +1,465 @@
...
+      neighbors:
+      - serverInterfaceRef:
+          name: host-01-if-01
+```
+
+These manifests configure a network roughly as described above: A
+spine connected to a leaf and that leaf connected to a host.
+
+The leaf also creates a VLAN around the interface towards the host
+to configure DHCP relay.
```
peanball, line 386: is this something we want to set explicitly or something that could be "derived" from the desire to have DHCP relay?

ieps/tbd-wire.md
```diff
@@ -0,0 +1,465 @@
...
+spec:
+  providerID: test://my-claimed-device
+  switchRef:
+    namespace: my-switch-namespace
+    name: my-switch-name
+    uid: my-switch-uid
+```
+
+By referencing the `Switch` back from the `Device`, we ensure that there can ever
+only be at most one `Switch` on a `Device`.
```
peanball, line 422: If we set up the desired topology by hand anyway, we could equally create `Switch` resources directly. Considering that the existence of a Server is handled by MetalAPI, and the network link configuration of a host is not controlled by WireAPI but instead observed and reported as state, we could also forego the concept of `Device` as overarching "catch-all" thing.

The things currently represented by `Device` are wildly different and might not benefit much by a common supertype.

peanball (commented):
Some comments.

ieps/tbd-wire.md
hardikdr: That sounds interesting. Do you have an example of how the formula could possibly look?

ieps/tbd-wire.md
hardikdr: I second this; it's much better to use a process to reduce reconfiguration than to enforce it via immutability.

ieps/tbd-wire.md
```diff
@@ -0,0 +1,465 @@
...
+  bgp:
+    asn: 0002
+    peerGroups:
+    - name: spines
+      neighbors:
+      - interfaceRef:
+          name: leaf-01-if-01
+    - name: leafs
+      neighbors:
+      - vlan: 1000
```
Giluerre, line 355: I probably understand what you mean, but I'll ask to be sure, by vlan you mean svi interface?

ieps/tbd-wire.md
afritzler: I think it is fair enough to start from the current state of affairs. That way we can keep the approach focused on the current fabric setup.

ieps/tbd-wire.md
afritzler: You are right. We might want to drop the `Server` part from this proposal and maybe outline this as a future enhancement to this proposal.

ieps/tbd-wire.md
afritzler: The `Link` resource is meant to describe the topology of devices. As the discovery process of devices is not part of this proposal, those resources are necessary to determine which port/interface is connected to whom.

ieps/tbd-wire.md
afritzler: The immutability aspect of the `Switch` object is the core concept of this proposal. It mimics the `Pod`s immutability in k8s which make the contract super explicit. In future this allow us to have a clear path on how to do a configuration rollout to a fleet of devices e.g. by having a Deployment/Replicaset types on top of `Switche`s.

ieps/tbd-wire.md
afritzler: See my comment here: https://github.com/ironcore-dev/enhancements/pull/21#discussion_r3507216675

ieps/tbd-wire.md
afritzler: See my comment here: https://github.com/ironcore-dev/enhancements/pull/21#discussion_r3507178694 The idea of the `Server` is to have this type in its own API group. So we should not confuse it with the `Server` resource coming from the `metal-operator` project.

ieps/tbd-wire.md
afritzler: @peanball can you elaborate on that a bit? Can you give an example of such a different configuration version?

ieps/tbd-wire.md
afritzler: Yes, you are right. The spec fields btw are debatable. They are currently derived from our ztp templates: https://github.com/ironcore-dev/sonic-operator/tree/main/internal/ztp/templates

ieps/tbd-wire.md
afritzler: This proposal is just the start of the journey. As described here: https://github.com/ironcore-dev/enhancements/pull/21#discussion_r3507216675 we can introduce high level concepts layer which will generate those immutable `Switch` objects.

ieps/tbd-wire.md
afritzler: You are right maybe we should sharpen the motivation of this proposal and what those `Switch` object should be used for.

ieps/tbd-wire.md
afritzler: Exactly this resource is just representing the physical reality.

adracus:
Superseded by #58 due to fork-based contribution mechanism.

ieps/tbd-wire.md
peanball: What I mostly meant is that each resource has a `Version`. Once that version changes you effectively have a different configuration, but for the same entity.

Main reasoning for this was that the same object remains, representing the same physical device, and there is no possibility for drift. Any references pointing to that object can remain as they are (e.g. links, etc.) and don't need to be recreated, which keeps the "undefined state" and possibly downtime to a minimum or rather eliminates it.

But further discussions made it clear that the "physical topology", so what is actually plugged in the data center, is going to be expressed by the operator and not auto-discovered. The physical device will be identifiable clearly enough to an old or new `Switch` object, of which there can only be one per physical switch. So in the end it doesn't make any difference if we create a new object or reconfigure (versioned) an existing one.

Finally, the `Link`s actually need to be shut down cleanly anyway when we take one of the switches out of rotation and reconfigure/reboot it. So it's good that we have to touch all referencing objects to make sure they've had the chance to reconfigure.

Checks for distinctness may be more complex with the immutable delete/recreate flow as you'll need to make sure that the object referencing the thing you are recreating is fully gone and replicated across the whole cluster. But that is also a solvable issue.

I understand that the immutability to some degree is also encouragement for the operators to not want to reconfigure things. There are also not many things to configure realistically. When they do, a "clean slate" approach is easier to verify as there cannot be any "stale" entries/references that someone forgot to touch.

ieps/tbd-wire.md
peanball: I think I finally fully understood the beauty of this: https://github.com/ironcore-dev/enhancements/pull/21#discussion_r3710515342

PR #29, opened 2026-06-11
IEP: Server firmware lifecycle full inventory, declarative rollout, scheduled maintenance

A-AKPINAR:
Proposes three extensions to existing resources: full firmware inventory (NIC/RAID/PSU) in `Server.statu`s, a rollout strategy + image checksum for the `*VersionSet` kinds, and an optional time window on `Se`rverMaintenance.

### DIFF
--- a/ieps/tbd-firmware-lifecycle.md
+++ b/ieps/tbd-firmware-lifecycle.md
@@ -1,0 +1,151 @@
+---
+title: Server firmware lifecycle - full inventory, declarative rollout, scheduled maintenance
+iep-number: NNNN
+creation-date: 2026-06-11
+status: implementable
+authors:
+  - "@A-AKPINAR"
+reviewers:
+  - TBD
+---
+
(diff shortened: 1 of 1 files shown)


ieps/tbd-firmware-lifecycle.md
```diff
@@ -0,0 +1,151 @@
...
+  strategy:                           # NEW: fleet rollout safety
+    maxUnavailable: 10%               # rollout limit across the selected fleet
+    verifyVersionAfterApply: true     # post-update version gate before continuing
+    rollbackOnFailure: true           # NEW: revert to the previous version if verification fails
+```
+
+When `verifyVersionAfterApply` fails, `rollbackOnFailure` reverts the affected server to its
+previously installed version before the rollout continues; that version is recorded in the
+resource status as the rollback target. The iDRAC retains the prior firmware image, which makes
+the revert possible.
```
horazont, line 117: This seems like a vendor-specific thing. Do all vendors retain the previous image? Or do we need to provide some kind of image cache to support this across vendors?

PR #30, opened 2026-06-11
Add Dell ServiceTag IEP

A-AKPINAR:
Surfaces the Dell `ServiceTag` (today only implicit in `sku/serialNumber`) as `Server.status.serviceTag` plus a selection label per the IEP-14 namespace.

### DIFF
--- a/ieps/tbd-dell-service-tag-iep.md
+++ b/ieps/tbd-dell-service-tag-iep.md
@@ -1,0 +1,71 @@
+---
+title: First-class Dell ServiceTag identifier
+iep-number: NNNN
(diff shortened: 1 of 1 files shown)


adracus (commented):
Hi @A-AKPINAR, thanks for the IEP, however, we do not plan to make vendor-specific fields part of the general `Server` spec. The mission is to keep `Server` vendor-agnostic. If you need a service tag field just for dell, why not implement a custom controller in your organization that watches `Server`s and e.g. writes an annotation or label to the server with the parsed service tag?

PR #38, opened 2026-06-22
Underlay encryption proposal

kitsudaiki:
Added proposal to encrypt the underlay network in order to encrypt the tenant traffic with IPsec between dpservice-instances.

### DIFF
--- a/ieps/NNNN-underlay-encryption.md
+++ b/ieps/NNNN-underlay-encryption.md
@@ -1,0 +1,220 @@
+---
+title: Underlay Encryption
+
+iep-number: NNNN
+
+creation-date: 2026-06-08
+
+status: implementable
+
+authors:
+
+- "@kitsudaiki"
+- "@mast-wch"
+- "@markus-hentsch"
+
+reviewers:
+
+- "tbd"
+
+---
+
+# IEP-NNNN: Underlay Encryption
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Alternatives](#alternatives)
+
+## Summary
+
+Implementation of an underlay encryption for the traffic between compute hosts. 
+This involves several networking components of IronCore such as
+dpservice and Metalnet.
+
+## Motivation
+
+Currently, the underlay network does not provide encryption. 
+However, to align with regulatory requirements for specific security domains (for example as mandated by the BSI), encryption is necessary. 
+Once enabled, all underlay traffic originating from Metalnet must be encrypted to ensure data confidentiality and integrity.
+
+### Goals
+
+- Encrypt IPv6-encapsulated traffic leaving the compute host using IPsec
+- Exchange keys for the IPsec encryption via IKEv2 between Metalnet instances
+
+### Non-Goals
+
+- Encryption per VNI is not in scope of this enhancement, because of the scalability. 
+  A Security Association (SA) can effectively only be established between 2 communication partners because of the sequence counter for replay protection. 
+  So even with encryption on host level instead of VNI level, this already will result in `2 * NUMBER_OF_HOST^2` SAs on each host in the worst case.
+  When also considering the VNI, the amount of SAs can easily explode in bigger deployments. The `2 *` comes from the fact 
+  that there is a SA for each direction, so each connection effectively creates 2 SAs.
+
+- The gRPC connection between Metalnet and dpservice is not encrypted additionally, because Metalnet and its connected dpservice are on the same host.
+
+- The option to offload encryption to the network card will be part of a subsequent enhancement.
+  Melanox BlueField network cards also support crypto offloading for IPsec (<https://docs.nvidia.com/doca/sdk/ipsec-crypto-offload/index.html>). 
+  This will require a secondary version of the implementation in the code, because the workflow between the CPU and the offloaded version is different. 
+
+## Proposal
+
+### Related open issues
+
+There are already 2 open issues related to this topic:
+
+- https://github.com/ironcore-dev/metalnet/issues/320
+- https://github.com/ironcore-dev/roadmap/issues/69
+
+The proposal for this enhancement document differs slightly from these issues and also goes more into technical details.
+
+### Overview:
+
+```mermaid
+graph TD
+    metalnet01["Metalnet"]
+    metalnet02["Metalnet"]
+
+    dpservice01["Dpservice"]
+    dpservice02["Dpservice"]
+
+    metalbond["Metalbond"]
+
+    subgraph Controller
+        ironcore-net
+        metalbond
+    end
+
+    subgraph Compute Host 01
+        metalnet01 --> |gRPC| dpservice01
+    end
+
+    subgraph Compute Host 02
+        metalnet02 --> |gRPC| dpservice02
+    end
+
+    metalbond <--> |TCP| metalnet02
+    metalbond <--> |TCP| metalnet01
+
+    metalnet02 <--> |IKEv2| metalnet01
+    dpservice02 <--> |IPsec| dpservice01
+    
+    ironcore-net --- |K8s Watcher| metalnet01
+    ironcore-net --- |K8s Watcher| metalnet02
+```
+
+### dpservice
+
+The DPDK graph in dpservice has to be extended with encryption and decryption nodes. 
+Those nodes shall use the IPSec implementation provided by DPDK (<https://doc.dpdk.org/guides/prog_guide/ipsec_lib.html>). 
+The library already provides the means for AES-256-GCM encryption/decryption as well as a SA database. 
+It also handles sequence numbers to protect against replay attacks. 
+The DPDK library will be configured to use ESP in transport mode for the single SAs. 
+As there is already encapsulation and decapsulation in place, this will basically result in tunnel mode ESP.
+
+Packets will be encrypted on the sender side after the IPv6-encapsulation and before the decapsulation on the receiver side. 
+This ensures, that only traffic leaving/entering the physical host, will be encrypted and decrypted to avoid unnecessary packet processing overhead. 
+An SA is created for each remote compute host, or rather remote dpservice instance, with which the local dpservice has to exchange packets. 
+These SAs are stored in the DPDK SA database. 
+The key for the established connection is provided by the Metalnet instance over the gRPC connection. 
+Whenever a new key is pushed by Metalnet, a new SA is created, even for target hosts, for which a SA already exist, in order to enable proper key rotation.
+
+### gRPC connection
+
+There are three endpoint changes necessary:
+
+- New fields will be added to the existing `CreateRoute` endpoint to encrypt the new connection right from the beginning:
+    - Authenticated Encryption with Associated Data (AEAD)
+        - 36 Byte key material, which also contains the salt
+    - Security Parameters Index (SPI)
+        - 32 bit identifier for the IKEv2/IPsec connection
+    - IP of the other communication partner for identification of the key
+
+- A new endpoint will be added to provide a key for key rotation.
+  This endpoint will be triggered every time when the IKEv2 in Metalnet shares a new key.
+  In case the related connection was previously unencrypted, which can be the case during a migration to this new version, 
+  then the connection will start becoming encrypted.
+
+- A new endpoint will be added to trigger the deletion of expired keys. Deletion shall happen after a short grace period with a brief overlap of old and new key to tolerate delayed traffic.
+
+### Metalnet
+
+The IPsec implementation of DPDK doesn't provide an implementation for IKEv2 (`Internet Key Exchange`), 
+but this special key exchange is highly required for IPsec by the RFC 
+`Cryptographic Algorithm Implementation Requirements and Usage Guidance for Encapsulating Security Payload (ESP) and Authentication Header (AH)` 
+(<https://datatracker.ietf.org/doc/html/rfc8221#section-3>). 
+To have better control over the management of the targets and to conform more closely to the general workflow in IronCore, 
+the key exchange will be done between Metalnet instances rather than between dpservices. 
+
+strongSwan, where IPsec and IKEv2 belongs to, 
+provides the Go library `govici` (<https://github.com/strongswan/govici>) to interact with the strongSwan library, which handles the IKEv2 process. 
+The strongSwan IKEv2 also includes automatic exchange of new keys after a certain amount of time and the connection of the key exchange is already encrypted. 
+So an additional distribution of a public key over the Metalbond is not necessary. 
+Furthermore the salt value for the encryption is also already generated and shared by the IKEv2 library.
+Only the address of the target is required. 
+
+Per default, the key exchanged by IKEv2 is stored in the vici module of the kernel. 
+Because the govici is only some kind of remote control, it can not intercept the process and catch the key before the kernel.
+In a regular case the IPsec would also run in the kernel itself and can access the key there. 
+Because DPDK runs in user-space and so does the IPsec in this proposal too, this workflow is not possible. 
+After exchanging a key with the library, Metalnet has to read the key from the kernel module to get it into the user-space, 
+for example with the netlink library (<https://github.com/vishvananda/netlink>) by a separate watch-loop, 
+and then push this key through the gRPC connection to its local dpservice instance.
+
+To avoid multiple IKEv2 runs for the same source-target-host-pair, Metalnet has to track the connections in a central config-map to avoid duplications.
+
+IKEv2 runs asynchronous. So when initially announcing a new route, 
+a wait function has to be added to wait until the IKEv2 has successfully exchanged keys, 
+to ensure that keys are available for the encryption and decryption on both hosts, 
+before Metalnet marks the connections ready for traffic.
+
+Because the encryption is done at the underlay traffic and the CRD's affect the overlay networks, 
+the new config to enable and disable the IKEv2 and IPsec will be added as a new `--enable-network-encryption` CLI flag to the Metalnet binary. 
+
+Following additional changes to the Metalnet environment are necessary:
+
+- For IKEv2 normally certificates are used to secure the connection. 
+  Every host needs its own certificate provided by the deploy process.
+
+- In case that the Metalnet container doesn't run in privilege mode, to be able to interact with the kernel in the IKEv2 process, the container needs additional capabilities:
+
+    ```
+    NET_ADMIN
+    NET_RAW
+    ```
+
+- Required system packages in the Metalnet pod:
+
+    ```
+    strongswan
+    strongswan-pki
+    strongswan-swanctl
+    libcharon-extra-plugins
+    iproute2
+    kmod
+    ```
+
+- For IKEv2 the `UDP` ports `500` and `4500` have to be open on the hosts.
+
+- Manifest of the Metalnet pods requires an update for the new CLI flag.
+
+## Alternatives
+
+- Use PSK (Pre-Shared-Key) instead of certificates for the IKEv2 process. This key then would have to be stored in a key manager like Vault, where all Metalnet hosts can access the key to be able to participate in the IKEv2 process. 
+
+- Custom implementations for encryption and key exchange:
+
+    - Create a custom key exchange mechanism between the Metalnet hosts based on TLS as alternative to IKEv2 key exchange.
+    - Add a custom encryption with AES256 in DPDK as an alternative to IPsec packet encryption.
+
+    Both would have a huge impact on the workload for implementation and maintenance and it would have a significant impact on security. Especially things like a BSI certification would be substantially harder with custom implementations for encryption and key exchange.
+
+- No dedicated enable/disable toggle per CLI, always-on encryption.
+
+- Instead of adding a new parameter to the initialize gRPC message, an enable flag, like for Metalnet, could be added to dpservice too, to enable encryption. This would require changes to the rollout process.
+
+- Instead of adding encryption as an extra node after encapsulation (and vice versa with decryption/decapsulation), the DPDK IPsec library's tunnel mode ESP could be explored and the nodes could be merged. This would mean to use NULL encryption algorithm (i.e. no encryption) as long as a route is supposed to be unencrypted. It would also mean not just to attach new nodes but change existing ones. 


trevex:
Ok, this will be a long comment, but as discussed let me outline some of the thoughts and discussion that went into https://github.com/ironcore-dev/metalnet/issues/320. Bare with me this was nearly two years ago and a lot of in-person ramblings back then, so bits might be incorrect and gaps are filled by my understanding of the original idea. So feel free to fact check me or propose changes. :)

If not anything, this might be input to consider for the current proposal.

> Disclaimer: Claude helped me paraphrase some section as my stream of conciousness was less readable. The text was proof-read and reflects my understanding.

## TL;DR

The concrete requirement was to at least encrypt per VNI. We want to separate tenants. Per hypervisor encryption does not give us anything a wireguard mesh wouldn't give us.
We want to improve the security and tenant separation. However per VNI introduce a bunch of unfortunate key exchange topics, so using a per vNIC with ephemeral keys and leveraging **metalbond for public key distribution** was the idea in a nut shell. (FYI: per VNI can also mean per Host + per VNI HKDF.)

Each host derives a symmetric key by ECDH and the resulting key drives **ESP/IPsec**. In theory this is supported by both DPDK key and the linux kernel via XFRM.

This avoids IKEv2's O(n^2) online pairwise handshakes and half-open-state DoS surface, costs zero data-plane round trips and reuses an asset IronCore already has - metalbond as key-distribution rendezvous. Similar to Cisco GETVPN/GDOI group keying.

> There is also kind of two options hidden in this comment. per host + per vNIC HKDF VS per vNIC. Originally the discussion in the past was per vNIK, but HKDF might be an interesting alternative. While typing I started leaning more towards per host + per vNIC HKDF :)

## The devil in the detail

However the details are important:

1. A key pair generated once at provisioning and kept for the vNIC's lifetime is **static/static ECDH**, *not* PFS. It provides forward secrecy only "proportional to the vNIC lifetime," which for long-running VMs is effectively none.
2. **Initialization vectors do not provide replay protection.** They provide nonce uniqueness (semantic security). Replay protection comes only from the ESP sequence number + receiver replay window.
3. **AES-GCM with control-plane static keys is explicitly prohibited by RFC 8221 §3**, because nonce uniqueness cannot be guaranteed across reboots — and Linux does not persist the outbound ESP sequence number across reboot, which causes catastrophic AES-GCM nonce reuse (breaks confidentiality *and* integrity).

## How do we intend to solve this?

We need to distribute *rotating ephemeral* DH public keys through metalbond instead of one fixed key. 

Each node rolls a node-level ephemeral on a configurable intervall; per-(VNI, route) traffic keys are derived by HKDF; (or straight per vNIC) each epoch installs a fresh ESP SA (new SPI) and the old SA is deleted after a make-before-break overlap.

This same mechanism **also fixes the AES-GCM nonce-reuse problem for free** (every epoch is a fresh key, so the sequence counter legitimately restarts), and it stays entirely within the constraints: **no IKEv2, no inline handshake, no new data-plane protocol, kernel-compatible** (it is exactly XFRM/`rte_ipsec` SA rekeying driven by the control plane instead of by an IKE daemon).

**One hard prerequisite:** metalbond is today **plaintext, unauthenticated TCP**. The entire scheme's resistance to MITM and to a rogue control plane depends on first giving metalbond peer authentication and a confidential channel (mutual TLS + peer-identity binding). Without that, public-key distribution is trivially MITM-able.

## Threat model

The proposal must defend against all four of the following:

| # | Adversary capability | Required defense |
|---|---|---|
| T1 | **Passive underlay eavesdropping** — reads any packet on the physical/underlay fabric. | Confidentiality (AEAD encryption). |
| T2 | **Active underlay MITM / injection / replay** — modifies, injects, or replays underlay packets. | Authenticated encryption (AEAD integrity) + anti-replay (ESP sequence + window). |
| T3 | **Malicious co-tenant / VNI breach** — one tenant tries to read or spoof another tenant's traffic. | Per-VNI key separation; no cross-VNI key reuse. |
| T4 | **Compromised host or rogue control-plane peer** — a compromised hypervisor or a rogue metalbond peer tries to impersonate, harvest keys, or MITM key distribution. | Authenticated, confidential control channel; peer-identity binding of announced keys; forward secrecy to bound damage of *past* traffic. |

**Important nuance on T4** We *assume the metalbond channel is trusted*. "Trusting the channel" therefore means **mutual TLS + peer-identity binding must be added to metalbond first**. Even then, a residual risk remains: a *legitimate but compromised* peer can still announce keys for the vNICs it genuinely hosts. Channel authentication does not prevent that; it only prevents *outsiders* and prevents a compromised peer from forging announcements for vNICs it does **not** host (because announcements are bound to the authenticated peer identity). Forward secrecy bounds how much *past* traffic a host compromise exposes.


## Rotating keys

DHE/ECDHE and the static design perform the **same Diffie–Hellman operation**. The only differences are *what gets published* and *whether the private key is discarded*. IKEv2 and WireGuard need an inline handshake **only because they have no other channel** to announce a fresh ephemeral public key. **IronCore already has that channel: metalbond.** Therefore we obtain DHE-grade forward secrecy by publishing *rotating* ephemeral public keys through metalbond and discarding old privates — no IKE, no inline handshake.

### Mechanism

Notation: `e_X^t` = node *X*'s ephemeral **private** key in epoch *t*; `E_X^t` = the matching **public** key. Each node runs the following independently and symmetrically.

**Per epoch, on each node *X*:**

1. **Generate** a fresh ephemeral DH keypair `(e_X^t, E_X^t)` — **in host memory only**, never persisted. One ephemeral per *node*, not per vNIC.
2. **Announce** the public key `E_X^t` over metalbond, alongside the routes the node already advertises, tagged with an epoch/validity field. The announcement is accepted by peers only for the prefixes/vNICs this authenticated peer is allowed to host.
3. **Learn** each peer's current ephemeral public key (`E_Y^t`) from the metalbond subscription it already maintains for the relevant VNIs.

**For each peer *Y* the node communicates with:**

4. **Agree** a per-node-pair master secret by ECDH:
   `master_XY^t = ECDH(e_X^t, E_Y^t) = ECDH(e_Y^t, E_X^t)` — both ends compute the identical value without any exchanged message beyond the published public keys.
5. **Derive** the per-`(VNI, route)` traffic keys from that master via HKDF, with the VNI in the info label so tenants never share a key:
   `key(VNI, route) = HKDF(master_XY^t, "ironcore-uenc-v1" ‖ VNI ‖ route-selector)`.
   (Optionally mix the node's long-term *identity* key into the HKDF to bind the ephemeral to an authenticated identity, WireGuard-style.)
6. **Install** an ESP SA with a **fresh SPI** for each `(VNI, route)` into the data path (XFRM on a kernel host, `rte_ipsec`/`rte_security` on a dpservice host), and attach it to the route entry so forwarding stays route-based and stateless (§8.3).

**On rotation to epoch *t+1* (driven by a time interval or a byte/packet ceiling, whichever comes first):**

7. Repeat steps 1–6 with a new ephemeral and a new SPI, **install the new SAs before shifting traffic**, switch egress, and only then **wipe the old ephemeral private `e_X^t`** and delete the old SAs after an overlap window. The lossless switch ordering is detailed in §7.7.

**Why this gives forward secrecy:** at epoch end both sides wipe `e^t`. A later host compromise yields only the *current* epoch's private key; past-epoch keys were derived from privates that no longer exist and cannot be recomputed. The FS window equals the rotation interval (WireGuard is likewise epoch-bounded — it rekeys roughly every 2 minutes).
- **Rotate one ephemeral per node, not per vNIC.** Per-VNI/route keys are derived by HKDF with the VNI as a label. This keeps control-plane churn O(nodes) while preserving per-VNI separation (T3).
- **The rotation knob is the whole spectrum.** Interval → ∞ degenerates to the static/static design (cheapest, FS ≈ vNIC lifetime). Interval = minutes ≈ DHE/PFS.



### Why this is kernel-compatible

Rotating keys = **installing a new SA (new SPI + key) per epoch and deleting the old after an overlap** — exactly what an IKE daemon does on `CREATE_CHILD_SA` rekey, just driven by metalnet. The kernel does not care who computed the key: `ip xfrm state add` with a fresh SPI is the same operation. The DPDK equivalent is an `rte_ipsec` SA replace. Both stacks hold multiple SAs simultaneously, so make-before-break works natively.

### Nonce and replay handling

- AES-256-GCM (RFC 4106), 16-octet ICV. Per-SA salt distributed with the key; per-packet 8-octet IV from the ESP sequence number.
- A **fresh key per epoch** guarantees nonce uniqueness even across reboots, satisfying RFC 8221: a node reboot **must force a new ephemeral (new epoch)**, never resume an old key. (Hard requirement.)
- **Anti-replay enabled** on the receiver (window ≥ 64; ESN/64-bit sequence to defer rollover). Because each epoch is a fresh SA, sequence restart at 1 is correct.

### The single make-or-break requirement

> **Ephemeral private keys live in host memory only. They are NEVER persisted to etcd / Kubernetes objects, and are wiped on rotation.**

### Estimated costs

- **Control-plane churn:** one rotating pubkey per node per epoch on metalbond, plus SA rekey churn scaling with `peers × VNIs × rotation_frequency`. Mitigate by tuning the interval (even hourly rotation beats months by orders of magnitude) and node-level ephemerals + HKDF.
- **Make-before-break overlap:** two SAs briefly live per pair to avoid drops during the switch.
- **Epoch synchronization:** propagation delay means the sender keeps the old SA until the peer has the new key (time-based overlap, or confirmation).
- Adds **no** new data-plane protocol, **no** inline handshake, **no** extra data-plane RTT, **no** IKEv2.

### Lossless rotation (make-before-break)

Rotation must not drop packets that are in flight while keys change. It does not, because **the receiver demultiplexes inbound ESP packets by the cleartext SPI**, and during a rotation both the old and new SA are alive at once. Every epoch gets a **new SPI** (never an in-place key swap), so a packet is always decryptable by whichever SA its SPI names — there is no "current key" the data path can race against.

**Rule 1 — install inbound before anyone sends on it.** A node installs the new *inbound* (decrypt) SA the moment the new key exists; traffic starts flowing on it only later. Because the key is `ECDH(my_ephemeral, peer_ephemeral)` and metalbond propagation delay is strictly positive, this is mostly automatic:

- **Rotation triggered by the receiver's own new ephemeral:** the receiver already holds the peer's current ephemeral, derives the new key, and installs its inbound SA *instantly* — before the peer even learns the new ephemeral. Safe by construction.
- **Rotation triggered by the sender's new ephemeral:** the sender already has the key but **must not start sending on the new SA until the receiver has installed it.** It defers the outbound switch by a conservative `T_switch ≥ max(metalbond propagation + SA-install latency)`, or — more robustly — until it observes a confirmation (a metalbond ack, or the first valid packet received from the peer on the new SPI).

**Rule 2 — delete the old inbound SA only after a quiet overlap (closes Gap B).** The sender stops using `SPI_t` the instant it activates `SPI_t+1`, but in-flight `SPI_t` packets are still arriving, so the receiver keeps `SPI_t`'s inbound SA for `T_delete ≥ T_switch + max in-flight/reorder time`. **Deletion is gated on the new SA being active plus the overlap elapsing — never on a blind wall-clock timer.** If metalbond is slow and the new key has not propagated, the old SA simply lives longer: the FS window stretches slightly, but **no packet is dropped.** 

**Why there are no secondary races:**

- **No sequence-number coordination.** Each SA has its own ESP sequence space (starting at 1) and its own anti-replay window. `SPI_t` at seq 5000 and `SPI_t+1` at seq 1 never confuse each other — a concrete advantage of new-SPI-per-epoch over in-place rekey.
- **Outbound switch is a single atomic SA-pointer flip.** A packet is encrypted under exactly one SA; packets already queued under the old SA stay decryptable (the receiver still holds the old inbound SA). No reordering-induced loss.
- **Per direction.** A→B and B→A are independent SAs; make-before-break applies to each independently.

**Backend specifics:**

- **Kernel/XFRM:** both SAs coexist in the SAD (distinct SPIs); `ip xfrm state add` the new one, flip egress preference to it, `ip xfrm state delete` the old one after the overlap.
- **DPDK `rte_ipsec` / inline offload:** the SADB (or, under `INLINE_PROTOCOL`, the NIC's SA table) holds both entries during overlap; inbound match is by SPI exactly as above. Under `INLINE_PROTOCOL` the NIC owns the sequence/replay state — do not double-manage it.

**The one real failure mode** is operator/implementation error: deleting the old SA too early (underestimating skew + in-flight time), or sending on a new SA before the receiver installed it. Both are precluded by (a) gating deletion on *new-SA-active + overlap* rather than a blind timer, and (b) the install-inbound-before-send ordering above.

---

## Architecture and data flow

### Components and responsibilities

```
            ┌──────────────────────── control plane ──────────────────────────┐
            │                                                                 │
  ┌─────────▼──────────┐    announce route + rotating pubkey    ┌─────────────▼─────────┐
  │  metalnet (host-1) │◄──────────────────────────────────────►│  metalbond reflector  │
  │  - gen identity kp │     subscribe VNI; learn peer pubkeys  │  (mTLS, peer identity)│
  │  - gen node ephem. │                                        │  carries: prefix,VNI, │
  │    (memory only)   │                                        │  next-hop, PUBKEY     │
  │  - ECDH + HKDF     │                                        └─────────────▲─────────┘
  │  - program SAs     │                                                      │
  └─────────┬──────────┘                                        ┌─────────────┴─────────┐
            │ gRPC: per-(VNI,route) SA                          │  metalnet (host-2)    │
            │ {SPI, key, salt, mode, lifetime}                  │  (symmetric)          │
  ┌─────────▼───────────┐                                       └─────────────┬─────────┘
  │ data plane backend  │                                                      │
  │  A) dpservice/DPDK  │  ◄══════════ encrypted ESP / AES-GCM ═══════════► (backend on host-2)
  │     rte_ipsec       │              over IPv6 underlay
  │  B) kernel XFRM     │
  └─────────────────────┘
```

- metalnet owns key agreement (ECDH + HKDF) and SA lifecycle; the data plane only enforces SAs it is given. This keeps crypto policy in one place and lets the same control logic drive either backend.

### Packet on the wire — keep IP-in-IPv6, add ESP transport mode

IronCore keeps its existing **IP-in-IPv6** encapsulation; ESP **transport mode** protects the encapsulated payload. This yields a **single outer IPv6 header** (no double encapsulation) and leaves dpservice's encapsulation model untouched:

```
 ┌──────────────────────────────────────────────────────────────────────────┐
 │ Outer IPv6 (host-1 → host-2, underlay)   next-hdr = ESP(50)              │
 │  ┌───────────────────────────────────────────────────────────────────┐   │
 │  │ ESP header: SPI, Seq                                              │   │
 │  │  ┌──────────────────────────────────────────────────────────────┐ │   │
 │  │  │ IV(8) │ ENCRYPTED{ inner overlay IP packet (VM-A→VM-B) } │ pad │   │
 │  │  └──────────────────────────────────────────────────────────────┘ │   │
 │  │ trailer(padlen,nexthdr=4/41)   ICV(16)                            │   │
 │  └───────────────────────────────────────────────────────────────────┘   │
 └──────────────────────────────────────────────────────────────────────────┘
```

- The **inner overlay IP packet — including tenant source/destination addresses — is inside the encrypted region**, so the underlay sees only host↔host. The same privacy as tunnel mode.
- Overhead ≈ ESP(8) + IV(8) + trailer + ICV(16) ≈ **~32–40 bytes**, *not* a second 40-byte IP header.

### Stateless, route-based (cryptokey-routing) — no constant tunnel

The design preserves dpservice's connectionless forwarding model — the WireGuard-style "cryptokey routing" operational model, **not** classic site-to-site tunnels:

```
 EGRESS (per packet, no flow state):
   dest overlay IP ─► route lookup (VNI) ─► { next-hop hostB IPv6, SPI, key }
                                            └─ ESP encrypt + IP-in-IPv6 encap ─► send to hostB
 INGRESS (per packet, no flow state):
   ESP SPI ─► SA lookup ─► decrypt ─► IP-in-IPv6 decap ─► VNI deliver
```

- The **only** retained state is the per-peer/route **SA** = `{key, SPI, replay window}`, attached as an *attribute of the route/next-hop entry* — the same per-peer state WireGuard keeps. There is **no connection state, no 5-tuple flow table, no tunnel up/down, and no handshake.**
- The SA is selected by *destination route + VNI*, not by a negotiated tunnel. Rotation swaps the SA attached to the route entry; the forwarding model is unchanged.
- "Stateless" here means *connectionless/route-based*, not "no key state" — no practical encryption scheme (WireGuard included) is stateless in the literal sense.

---

## Linux kernel (XFRM) compatibility

**Compatible.** Control-plane-keyed ESP over XFRM is well-trodden (Cilium, Calico, strongSwan route-based VPNs all program XFRM directly). It might look like this:

```bash
# --- SAs (epoch t uses reqid 1) ---
# Outbound  A->B   (SPI 0xa1)
ip xfrm state add src fc00:1::1 dst fc00:2::1 proto esp spi 0xa1 \
    mode transport reqid 1 \
    aead 'rfc4106(gcm(aes))' 0x<KEY_AB_t:32B><SALT_AB_t:4B> 128 \
    replay-window 64 flag esn
# Inbound   B->A   (SPI 0xb1)
ip xfrm state add src fc00:2::1 dst fc00:1::1 proto esp spi 0xb1 \
    mode transport reqid 1 \
    aead 'rfc4106(gcm(aes))' 0x<KEY_BA_t:32B><SALT_BA_t:4B> 128 \
    replay-window 64 flag esn

# --- Policies ---
# OUT: ipip packets (proto 4) A->B must get ESP transport; reqid 1 selects the epoch-t SA.
ip xfrm policy add src fc00:1::1 dst fc00:2::1 proto 4 dir out \
    tmpl src fc00:1::1 dst fc00:2::1 proto esp mode transport reqid 1
# IN: require ESP transport on ipip packets B->A. reqid omitted (=0, wildcard) so BOTH
#     epochs are accepted during overlap; inbound SA is chosen by SPI regardless.
ip xfrm policy add src fc00:2::1 dst fc00:1::1 proto 4 dir in \
    tmpl src fc00:2::1 dst fc00:1::1 proto esp mode transport
# (add matching dir fwd policy if the host forwards rather than terminates)
```

Host B is the mirror image: swap `src`/`dst` throughout, keep the **same per-direction keys**, same SPIs (`0xa1` is the A→B SA on both ends; `0xb1` is B→A on both ends).

Ordering matters: **install new inbound SAs first, then switch egress, then delete old.**

```bash
# ── Rule 1: install the NEW INBOUND SA before anyone sends on it ──
#    (run on EACH host for its own inbound direction; epoch t+1 uses reqid 2)
# Host A — new inbound B->A (SPI 0xb2):
ip xfrm state add src fc00:2::1 dst fc00:1::1 proto esp spi 0xb2 \
    mode transport reqid 2 \
    aead 'rfc4106(gcm(aes))' 0x<KEY_BA_t1:32B><SALT_BA_t1:4B> 128 \
    replay-window 64 flag esn
#    Host B symmetrically installs new inbound A->B (SPI 0xa2, reqid 2).
#    Now both 0xb1 and 0xb2 decrypt on A; both 0xa1 and 0xa2 decrypt on B (SPI-demuxed).

# ── Switch EGRESS only after the peer has its new inbound SA ──
#    (gate by T_switch ≥ max propagation, or by confirmation — §7.7)
# Host A — add new outbound A->B (SPI 0xa2, reqid 2):
ip xfrm state add src fc00:1::1 dst fc00:2::1 proto esp spi 0xa2 \
    mode transport reqid 2 \
    aead 'rfc4106(gcm(aes))' 0x<KEY_AB_t1:32B><SALT_AB_t1:4B> 128 \
    replay-window 64 flag esn
# Atomically repoint the OUT policy to reqid 2 → egress flips to the new SA in one step:
ip xfrm policy update src fc00:1::1 dst fc00:2::1 proto 4 dir out \
    tmpl src fc00:1::1 dst fc00:2::1 proto esp mode transport reqid 2

# ── Rule 2: delete OLD SAs only after the overlap window (T_delete) ──
#    By now no epoch-t packets remain in flight. Run on both hosts.
ip xfrm state delete src fc00:1::1 dst fc00:2::1 proto esp spi 0xa1   # old out A->B
ip xfrm state delete src fc00:2::1 dst fc00:1::1 proto esp spi 0xb1   # old in  B->A
```


## Open questions

1. **Identity granularity:** per-node or per-vNIC identity keys? Per-node is far less churn; per-vNIC gives finer authentication and natural regeneration on reschedule. Personally torn a bit on this, but per node identity + per VNI HKDF might be best trade off...
2. **Rotation cadence default:** time-based (e.g. 1 h) vs byte/packet-driven vs both. What FS window does IronCore want as the default?
3. **make-before-break overlap policy:** conservative time-based `T_switch`/`T_delete` vs explicit peer-has-new-key confirmation (metalbond ack or first-packet-on-new-SPI). Time-based is simpler; confirmation-based is robust to propagation-delay outliers.
5. **MTU strategy:** ESP overhead on top of IP-in-IPv6; path-MTU vs fixed reduction; fragmentation handling on inline-offload NICs.
6. **South-north / NAT / load-balancer-target routes:** how keys apply to `NAT` and `LOADBALANCER_TARGET` next-hop types, and to traffic leaving the overlay.
7. **Root of trust for identity keys:** out of scope now (no KMS/HSM), but where does long-term identity authentication ultimately anchor?

ieps/NNNN-underlay-encryption.md
```diff
@@ -0,0 +1,220 @@
...
+the key exchange will be done between Metalnet instances rather than between dpservices. 
+
+strongSwan, where IPsec and IKEv2 belongs to, 
+provides the Go library `govici` (<https://github.com/strongswan/govici>) to interact with the strongSwan library, which handles the IKEv2 process. 
+The strongSwan IKEv2 also includes automatic exchange of new keys after a certain amount of time and the connection of the key exchange is already encrypted. 
+So an additional distribution of a public key over the Metalbond is not necessary. 
+Furthermore the salt value for the encryption is also already generated and shared by the IKEv2 library.
+Only the address of the target is required. 
+
+Per default, the key exchanged by IKEv2 is stored in the vici module of the kernel.
```
trevex, line 160: nitpick: vici is strongSwan's userspace management socket. SA is stored in kernel XFRM state.

ieps/NNNN-underlay-encryption.md
```diff
@@ -0,0 +1,220 @@
...
+    ```
+    strongswan
+    strongswan-pki
+    strongswan-swanctl
+    libcharon-extra-plugins
+    iproute2
+    kmod
+    ```
+
+- For IKEv2 the `UDP` ports `500` and `4500` have to be open on the hosts.
```
trevex, line 201: While the host has the PF (due to switchdev) we need to make sure dpservice's eSwitch steering does not catch the IKE packets.

ieps/NNNN-underlay-encryption.md
```diff
@@ -0,0 +1,220 @@
...
+### Goals
+
+- Encrypt IPv6-encapsulated traffic leaving the compute host using IPsec
+- Exchange keys for the IPsec encryption via IKEv2 between Metalnet instances
+
+### Non-Goals
+
+- Encryption per VNI is not in scope of this enhancement, because of the scalability. 
+  A Security Association (SA) can effectively only be established between 2 communication partners because of the sequence counter for replay protection. 
+  So even with encryption on host level instead of VNI level, this already will result in `2 * NUMBER_OF_HOST^2` SAs on each host in the worst case.
```
trevex, line 54: Slightly misleading:
Host-level SAs give each host `2·(N−1) ≈ O(N)` SAs (in+out to each peer); the `≈2N²` figure is the **fleet-wide** total, not per-host.

ieps/NNNN-underlay-encryption.md
```diff
@@ -0,0 +1,220 @@
...
+Once enabled, all underlay traffic originating from Metalnet must be encrypted to ensure data confidentiality and integrity.
+
+### Goals
+
+- Encrypt IPv6-encapsulated traffic leaving the compute host using IPsec
+- Exchange keys for the IPsec encryption via IKEv2 between Metalnet instances
+
+### Non-Goals
+
+- Encryption per VNI is not in scope of this enhancement, because of the scalability.
```
trevex, line 52: Is the goal just encryption or improving tenant separation and security? I would argue we need some level of per VNI encryption. Using HKDF might be an option here as well to stay with per Host IKE, but derive multiple keys.

ieps/NNNN-underlay-encryption.md
```diff
@@ -0,0 +1,220 @@
...
+  then the connection will start becoming encrypted.
+
+- A new endpoint will be added to trigger the deletion of expired keys. Deletion shall happen after a short grace period with a brief overlap of old and new key to tolerate delayed traffic.
+
+### Metalnet
+
+The IPsec implementation of DPDK doesn't provide an implementation for IKEv2 (`Internet Key Exchange`), 
+but this special key exchange is highly required for IPsec by the RFC 
+`Cryptographic Algorithm Implementation Requirements and Usage Guidance for Encapsulating Security Payload (ESP) and Authentication Header (AH)` 
+(<https://datatracker.ietf.org/doc/html/rfc8221#section-3>).
```
trevex, line 149: From my understanding RFC8221 does not require IKEv2. RFC 8221 §3 actually mandates **automated key management** for AES-GCM (because manual keying can't guarantee nonce uniqueness across reboots). The RFC compels *an automated KMS*, of which IKEv2 is the canonical, but not the only option.

trevex (commented):
The IEP is a **coherent, deployable design built on standard, certifiable components** (IKEv2 + strongSwan + kernel IPsec). That is its biggest strength and is well-aligned with its stated drive: BSI regulatory compliance. It is **not wrong in its crypto choices** (AES-256-GCM, ESP, automated keying). But there are several nuances I pointed out in comments.

Compared to the design lined out in my earlier comment: they are not strictly better/worse — they optimise for different things. The IEP optimises for **certifiability and standard tooling**; the ECDH design optimises for **tenant separation, scale, and avoiding fragile kernel↔userspace key plumbing**. The primary difference is that the IEP takes a connection-oriented key establishment with setup latency and O(n^2) scaling but provides standard, certifiable auth via IKEv2.

Either way, when using IKEv2 I would propose a hybrid also drawing inspiration from both and I would argue we should consider the following points:
1. Can we improve tenant separation by introducing a HKDF-based approach (outlined in my earlier comment)?
2. The kernal SAs and netlink scraping are finicky and potentially fragile, we should consider userspace IKE. This should be possible via strongSwan plugin or using vici's event API. Or by implementing IKE in metalnet directly...

With those two points addressed I would argue this is a solid option. We need to be concious about the O(n^2) scaling though: A 1.000 nodes means 500.000 IKEv2 sessions (`N*(N-1)/2`).

trevex:
To finish it off a quick comparison of current state (which would change by addressing the proposed changes):

| Topic | ECDH / prev. comment | This IEP |
|---|---|---|
| **Key agreement** | ECDH over **rotating public keys distributed by metalbond**; **0 data-plane RTT** | **IKEv2** online handshake between Metalnet pairs (UDP 500/4500); 2-RTT + half-open state |
| **Per-VNI tenant separation** | **Yes** — per-`(VNI,route)` keys via HKDF | **No** — host-level SAs only; one key per host-pair across all tenants, **but addressable as commented** |
| **Forward secrecy** | Tunable via epoch rotation (none → PFS) | **Yes** — IKEv2 DH rekey (strongSwan auto-rekey) |
| **Peer authentication** | metalbond **mTLS + identity binding (must be added)** | **IKEv2 certificates per host** — mature, standard |
| **RFC 8221 compliance** | Yes — automated keying via fresh key+SPI per epoch | Yes — via IKEv2 |
| **How dpservice (DPDK) gets keys** | Derived in Metalnet userspace, pushed over gRPC — **no kernel involvement** | strongSwan negotiates **in kernel**, Metalnet **scrapes via netlink**, pushes over gRPC — potentially fragile **but addressable as commented** |
| **Works under dpservice's switchdev NIC model?** | **Yes** — control plane is metalbond TCP / can ride a mgmt path; data keys never need a kernel netdev | **Feasible** — switchdev keeps the PF in the kernel, so IKE (UDP 500/4500) can run; conditional on underlay addressing on the kernel PF + eSwitch not capturing IKE + rekey-scraping machinery. Would be impossible only under vfio-pci, which dpservice does not use. |
| **Scalability** | O(n) pubkey distribution; no online handshakes | O(n²) IKE sessions fleet-wide; central config-map to dedupe; half-open state |
| **Standard components / certifiability (BSI)** | Custom ECDH-over-metalbond — **harder to certify** | **Standard IKEv2/strongSwan** — **easier to certify** |
| **Operational footprint** | metalbond field + HKDF + SA programming | strongSwan + govici + netlink scraping + per-host certs + extra capabilities/packages |
| **Connection model** | **Connectionless** key establishment — no handshake; route usable as soon as peer pubkey is distributed | Data plane connectionless, but **connection-oriented key establishment** — IKE SA per host-pair, handshake-gated readiness, DPD liveness, connection tracking |
| **Make-before-break** | Explicit SPI-demux ordering | Present in spirit (new SA per key + overlap) |

trevex:
There is something that was not discussed in regards to this PR:
How does the proposed design deal with PQC?

The answer in regards to strongSwan is straight-forward: PQC supported.

The downside is that IKEv2 gets even noisier on the packet level as PQC introduce fragmentation and additional rounds for intermediates.

However the original idea I outlined also loses a lot of its strengths. There is no PQC non-interactive key exchange such as DH, which means we need to explicitly exchange something between pairs of nodes. (Well there is CSIDH but is is not standardized, contested by researchers and slow. Related KEM SIDH/SIKE was *completely broken* in 2022.)

If we apply Kyber/ML-KEM to the metalbond option the primary advantage is the star topology rather than the full-mesh IKE. Essentially increasing scalability similar to dynamic Spoke-to-Spoke Meshes (DMVPN / ADVPN). We would have to exchange ML-KEM ciphertext via metalbond as well. So in the worst O(n^2) ciphertexts, but not the protocol overhead of IKE, e.g. O(n^2) connections with DPD et cetera.

**There is also a third option:** Next-gen SD WAN often handles these use-cases through a centralized controller. Metalbond (or a new controller) genereates keys and "pushes them down" to the nodes.

Other options to consider might also be MLS/TreeKEM or hybrid approaches such as central controller for PSKs.

Another interesting thought is leveraging a layed hybrid approach. Projects such as https://github.com/rosenpass/rosenpass use ML-KEM to exchange PSKs for Wireguard. In a nutshell if we encrypt public keys, DH becomes PQC resilient. 

The BSI explicitly mandates these hybrid approaches ("defense in depth") in some cases, because:

| | vs classical attacker | vs quantum attacker | vs an ML-KEM flaw/bug |
|---|---|---|---|
| X25519-only (today) | safe | **broken** | n/a |
| ML-KEM-only (PQC-only) | safe | safe | **broken** |
| **Hybrid (X25519 + ML-KEM)** | safe | safe | safe |

The aforementioned KEM SIDH/SIKE incident is a good example for why hybrid might be worthwhile. It is however important to note that BSI prefers Hybdrid and NIST advocates for PQC only as they believe less complexity is better.

A hybrid mechanism for the metalbond approach might look as follows:

1. Generate two ephemeral keypairs, memory-only: an X25519 pair `(e_X, E_X)` and an **ML-KEM-768** pair `(sk_X, pk_X)`.
2. Announce both public parts `E_X` (32 B) and `pk_X` (1184 B) over metalbond, tagged with epoch + algorithm IDs.
3. For each peer *Y*, once *X* has learned *Y*'s `(E_Y, pk_Y)`:
   - X25519 leg (non-interactive): `ss_dh = DH(e_X, E_Y)` — derivable immediately, both sides, no message.
   - ML-KEM leg (needs a ciphertext): a deterministic rule picks the **encapsulator** — e.g. the node with the lexicographically-lower underlay IPv6 address. Say *A* encapsulates to *B*: `A: Encapsulate(pk_B) -> (ct_AB, ss_kem)`, and *A* publishes `ct_AB` over metalbond addressed to the pair/epoch. `B: Decapsulate(sk_B, ct_AB) -> ss_kem`.
4. Combine with a KDF that binds both secrets and the transcript (X-Wing-style):
   `master = KDF(ss_kem ‖ ss_dh ‖ ct_AB ‖ pk_B ‖ label)`.
   Security holds if the KDF is sound **and** (ML-KEM **or** X25519 is unbroken).
5. Derive per-`(VNI, route)` keys: `key(VNI,route)

trevex:
I mentioned MLS/TreeKEM before. MLS is RFC9420 and has decent conformance tests and other implementations, so I went ahead and wanted to see whether MLS is a good fit here. It scales  much more gracefully than IKEv2.

Rather than theorize I drafted a spec and implemented a prototype with Claude:
https://github.com/trevex/mls-go

What I did:

I used **MLS (RFC 9420)** as the group-key-agreement primitive — it *is* the standardized,
interop-tested version of "publish rotating ephemerals, derive traffic keys, rekey with
make-before-break", and it gives forward secrecy + post-compromise security and group
keying for free.

More importantly it is a well-established standard and not something hand-rolled.

What exists today (as an exploration, not a product):

- **`mls-go`** — a zero-dependency (stdlib-only) Go MLS implementation. 15/15 official
  RFC 9420 KATs pass; **interop-validated against OpenMLS** on suite 1, both plaintext
  and encrypted handshakes. Includes a **post-quantum** hybrid suite (X-Wing =
  X25519 + ML-KEM-768) alongside the classical suites.
- **ironcore integration layer** — maps **a VNI ↔ an MLS group**, derives the **ESP
  AES-256-GCM SA** (key+salt+SPI) from the epoch's exporter secret, does
  **make-before-break** (epoch low-byte in the SPI so old/new SAs coexist and demux),
  SPIFFE/PKI credentials, and external-commit recovery.
- **deterministic metalnet/metalbond simulation** — single-seeded discrete-event model
  (no goroutines, reproducible) driving the *real* crypto: 2 reflectors + N hosts + M
  VNIs under packet drops, reflector-down, and partition. Headline result:
  **dual-group redundancy ⇒ zero tenant data-plane packet loss** across reflector
  failover/partition; a negative control deliberately fails the zero-loss check to prove
  it has teeth.

### How it maps onto this proposal's threat model

- **T1/T2** — identical to the IKEv2 proposal: ESP AES-256-GCM AEAD + anti-replay. Same
  data path (XFRM / `rte_ipsec`); MLS only replaces the *key agreement*, not the wire.
- **T3 (tenant separation)** — a group **per VNI** ⇒ keys are never shared across tenants
  by construction. This is the per-VNI separation I argued for; it does **not** require
  the per-host-only model that this IEP lists as a non-goal.
- **T4 (rogue/compromised control plane)** — two wins. MLS epochs give the rotating-key
  forward secrecy directly (a commit = a new epoch = a fresh ESP SA). And **"blind
  metalbond"**: member handshakes can be framed as MLS `PrivateMessage`, so the reflector
  **relays only ciphertext** — it can't read VNI membership churn, identities, or leaf
  keys. That meaningfully lowers what we must trust metalbond's *channel* with (mTLS is
  still needed for ordering integrity / peer-identity binding, but no longer for
  confidentiality of who-is-in-which-VNI). Validated cross-stack against OpenMLS.
- **RFC 8221 nonce-reuse** — solved the same way as my sketch: every epoch is a fresh
  key, so the ESP sequence legitimately restarts; a reboot just forces a new epoch.

### Benefits

- **Blind metalbond** — DS sees opaque, ordered ciphertext only.
- **No IKEv2, no inline handshake, no O(n²) online pairwise key exchange, no new
  data-plane protocol** — reuses metalbond as the distribution channel, exactly as argued.
- **Standardized + interop-tested + PQ-ready** — not a bespoke crypto scheme we have to
  certify alone; OpenMLS-validated; X-Wing gives a post-quantum option now.
- **Lossless rekey/failover is the part that's actually validated** in the sim, not just
  asserted.

### Tradeoffs

- **MLS is heavier than a thin pairwise ECDH+HKDF.** It buys group keying, FS/PCS,
  membership, and a standard — worth it for *per-VNI group* keys; arguably overkill if the
  requirement is only per-host pairwise keys.
- **Ordering.** MLS commits need a single linearization point per group; metalbond is not
  a consensus system. The sim uses **dual-group redundancy** (each reflector orders its
  own replica, no coordination) → **~2× control-plane + SA state** in exchange for zero
  reflector-to-reflector coupling and zero packet loss. That's the scaling cost.
- **Churn** scales with VNI count × membership-change rate (VM migration = group churn);
  mitigated by node-level rekey + HKDF-per-VNI, but still more than per-host-only.
- **Metalbond auth is still a prerequisite** — blind handshakes reduce the confidentiality
  bar on the channel but a *legitimate-but-compromised* peer can still announce for VNIs it
  hosts; peer-identity binding remains required (same conclusion as my earlier comment).
- **Maturity** — this is a tested library + a simulation *model*, not a metalnet/dpservice
  integration. The DPDK/XFRM SA programming is the same as this IEP targets; the metalnet
  wiring is not built.

### Is it a valid option going forward?

Yes — as a concrete, de-risked realization of the rotating-key-over-metalbond direction,
and complementary to (not necessarily competing with) the IKEv2 path. It's the strongest
fit when **per-VNI separation, forward secrecy, post-quantum readiness, and minimizing
trust in metalbond** are priorities. It can also be adopted **incrementally**: use MLS
purely as the group key-agreement/distribution mechanism feeding the *same* ESP data path
this IEP already designs (rte_ipsec / XFRM), without committing to the rest up front. The
simulation suggests the operationally scary part — lossless rekey and reflector failover —
holds.

trevex:
@afritzler raised concerns how deeply and tightly coupled this effort should be.

I would propose we focus in the next community calls on:
* Which ironcore bits (primarily dpservice) need to be extended
* Which additional (addon) components are introduced and how do they interact with ironcore (core) components? Which APIs or "plugin points" are required?

guvenc:
@trevex This will be also a long comment :)
I have been following the discussion for a while and I would like to bring in my perspective to the discussed points as well.

I have prepared a diagram so that I can use it explain my thoughts better.

image

In order to set the stage, we would need first trust anchors for the metalbond clients to talk to the metalbond server. We have actually two types of metalbond clients, the one which is embedded into the metalnet kubernetes controller and the one which is running as a standalone metalbond client on a router. 

I think there is already an agreement that the clients and the metalbond server would need mTLS to mutually authenticate each other and they will need also certificates and a PKI infrastructure to make the mTLS work.  I would propose to use [cert-manager](https://cert-manager.io/) with an internal CA to make PKI control plane work, nicely integrated to the Kubernetes machinery.
One thing to note here is that metalbond server is not installed as a kubernetes pod in current IronCore but this can be easily adjusted.

Another architectural point to emphasize is that the mentioned metalbond client types use netlink and dpservice gRPC for route installation so the backends are interchangeable for route installation,  the similar backend pattern can be used for IPSec as well, so that dpservice can do the IPSec encryption with gRPC + rte_ipsec and the Linux Kernel backend can use netlink + XFRM and metalbond core logic can stay unchanged for both use cases.

One of the open questions here is also how to encrypt the south-north traffic. There is already the tendency of using per VNI based keys to encrypt the east-west traffic but south-north traffic comes with several questions:

- Do we need to encrypt south-north traffic at all, as this traffic is not destined to live in the datacenter and going to/coming from internet.
- If we want to encrypt. One idea would be using "Public VNI 100" key to encrypt the south-north traffic, which is used by the metalbond client on the router to install the public facing routes. If we choose this approach, every VNI would need to know also VNI 100's key.
   - south-north traffic is easy to determine to be encrypted by VNI 100's key in dpservice, as it is normally the default route traffic going to the router but for the returning traffic from internet, would the router need to identify the target VNI and encrypt with that key or would the router always use VNI 100 key also for the return traffic ?
   - LOADBALANCER and NAT routes are also special cases as when they are coming from internet to the router, they can first hit one dpservice and hit another one based on the routing table of the first dpservice. 
First router -> dpservice1 traffic can be either encrypted in VNI 100 or target VNI and the second dpservice1 -> dpservice2 east-west bounce would need the target VNI's encryption key. 
 
Most of the discussion here was also about the key exchange, forward secrecy and key rotation. I see three ideas discussed so far:
- Each metalnet client can do a mesh with another metalnet client for key exchange and there will be dpservice-to-dpservice IPsec tunnels. (To be honest, as being the former maintainer of the ironcore networking components, this approach seems to be the least suitable to the way current networking components work)
- ECDH - Rotating pubkeys via metalbond, Per-VNI keys via HKDF.
- MLS - RFC 9420 group keys, 1 group per VNI, PQ-ready. (This is my favourite approach at the moment. @trevex has even a working prototype)

Another point I want to also mention is that the performance of the rte_ipsec and XFRM needs to be tested. I would strongly argue that we need to build PoCs to test the both cases. From feasibility perspective and from the performance perspective.

Last thing I would like to mention is also the question which IronCore interfaces might be effected from underlay encryption. My current guess based on the design discussion here:
- dpservice gRPC interface.
- metalnet NetworkInterface, LoadBalancer Kubernetes objects.
- metalbond would get mTLS.
- IronCore API probably in ironcore-net would need to mark the network related API resources as "encrypted" so that underlaying machinery can do the encryption job. Probably NetworkInterface, Loadbalancer and NATGateways can be created with "encrypted" flag on IronCore level. (Or some sort of signal from the higher IronCore API)

In a nutshell:
- Initial Trust Anchor through cluster internal CA with the help of [cert-manager](https://cert-manager.io/)
- Metalbond client and server get mTLS support.
- dpservice + GRPC + [rte_ipsec ](https://doc.dpdk.org/guides/prog_guide/ipsec_lib.html) / linux kernel + XFRM PoC feasibility and performance check.
- Decision about north-south traffic encryption. (Potential solution suggested above in my text - VNI 100 Usage)
- Decision about how to handle  Loadbalancer and NAT (Potential solution suggested above in my text)
- Decision about what to use for key-exchange, forward secrecy and per VNI encryption. (MLS looks promising)

trevex:
Let me emphasize a couple of points @guvenc made and propose answers to some others:
* While I believe north-south traffic does not necessarily need encryption, we can leave this open and first focus on east-west encryption.
* We should test performance, but always make sure that encryption is optional as there will be a performance impact regardless

I would also propose the following guiding principle:
> **All MLS / crypto / group-membership logic lives in a separable addon. The
> core components (IronCore API, metalnet, dpservice, metalbond) gain only
> narrow, encryption-agnostic seams.**

I believe the changes apart from implementing encryption in dpservice and adding a "SA install/rotate/remove API" are most likely minimal. Other components that a relevant would be:
1. metalnet should receive an "encryption ready" status condition, so that starting the VM can be "gated", if encryption is enabled.
2. A minimal spec change to Network and/or NetworkInterface to include "encrypted".

With these available an independent set of components could drive the key management and just install SAs as necessary.

### Some additional concerns/thoughts

Regardless of MLS or IKEv2 there is one central issues with scaling encryption that was pointed out to me:
Inline crypto offloading in DPDK (and presumably also XFRM) is limited to 128 SAs.

#### Here is a catch in regards to MLS:
While in theory MLS can bring us to 128 SAs. If SAs are shared across groups, this means up to 128 Networks per Host can be accelerated. However there is a problem...

MLS gives a VNI a **group key** shared by all M members. Two issues arise when
that key drives ESP:

1. **AES-GCM nonce reuse**: two senders using the same key+nonce
   → forgery / key recovery. **Already solved** by per-sender GCM salt.
4. **ESP anti-replay window collision**: RFC 4303's replay window is
   **per-SA (per-SPI)**. Today **one group SPI**
   shared by all senders, so N senders each starting at sequence 1 collide in the
   receiver's single window → legitimate packets dropped as replays. It also
   breaks receive-side demux: with one SPI the receiver can't tell which sender's
   salt to use.

The solution: per-sender SPI (the GDOI / G-IKEv2 group-SA model)

Give **each sender its own SPI** so each gets its own key stream *and* its own
anti-replay window. Key material stays group-shared; only SPI and salt become
per-sender. Each receiver installs **M inbound SAs per VNI** (same `K_group`,
distinct SPI+salt per sender leaf), each with an independent RFC 4303 window.

#### Back were we started:

However now we end up with too many SAs to fully offload everything...
How much the CPU is hit we need to test properly, but an [older Xeon did 1.89Tbps](https://builders.intel.com/docs/networkbuilders/intel-avx-512-high-performance-ipsec-with-intel-xeon-scalable-processor-technology-guide-1683018859.pdf), so a guesstimate would certainly still be multiple cores...
 
There are alternatives (such as Time-Based Anti-Replay), but they would break XFRM compatibility.  

Regardless I would argue we should add the IPsec support to dpservice and add the necessary extension points to be able to start testing and iterating key exchange and more.

kitsudaiki:
As decided in the community-meeting, this PR will be closed and replaced by the two smaller PR's

- https://github.com/ironcore-dev/enhancements/pull/50
- https://github.com/ironcore-dev/enhancements/pull/51

Based on the results of the comments of this PR here and the discussions in the community meetings, the new key exchange proposal uses MLS instead of IKEv2

PR #39, opened 2026-06-23
BPF dataplane proposal as Cilium-inspired alternative

trevex:
Added a proposal for a eBPF-based dataplane sparked by a recent discussion.

Looking for feedback and an open discussion in regards to the future developments of the dataplane and options to better support fog/edge needs as well.

### DIFF
--- a/ieps/21-ebpf-tc-dataplane.md
+++ b/ieps/21-ebpf-tc-dataplane.md
@@ -1,0 +1,332 @@
+---
+title: An eBPF (XDP + tc-BPF) Dataplane for dpservice — a Cilium-inspired alternative
+
+iep-number: 21
+
+creation-date: 2026-06-23
+
+status: draft
+
+authors:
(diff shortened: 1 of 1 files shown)


maxmoehl:
What you present sounds very intriguing. Unfortunately, after some discussion with @afritzler @FlorinPeter and @balpert89 we came to the conclusion that the IronCore project should rather not take on alternative implementations of such core networking components. Since dpservice also offers a software-mode the benefit of maintaining two completely separate stacks is limited compared to the maintenance overhead.

We encourage everyone to extend our eco-system by developing drop-in components or extending our APIs, but we don't have the capacity to take on such projects into the core project. Should the implementation mature we would be happy to feature it on our home-page. One note on the gRPC API you've implemented against: this API is not considered public and we make no guarantees when it comes to the stability, of course we will still do our best to keep it stable.

PR #41, opened 2026-06-24
`Machine` Eviction Proposal

### LINKED ISSUE #40: `Machine` Eviction Proposal
- [ ] Research K8s Cordon & Eviction concept and implementation
- [ ] Write proposal

lukasfrank:
Added IEP-21: MachinePool Eviction

### DIFF
--- a/ieps/21-machine-eviction.md
+++ b/ieps/21-machine-eviction.md
@@ -36,7 +36,7 @@ reviewers:
 
 This IEP defines how `Machines` are evicted from a MachinePool.
 
-We introduce a `NoExecute` taint effect. When set on a `MachinePool`, a taint eviction controller issues DELETE on all bound `Machine`s. 
+We introduce a `NoExecute` taint effect. When set on a `MachinePool`, a taint eviction controller issues DELETE on every bound `Machine` that does not tolerate the taint. 
 The provider handles graceful VM shutdown via its existing finalizer before the object is fully removed.
 
 ## Motivation
@@ -47,14 +47,13 @@ There is no mechanism to gracefully shut down VMs in an ordered fashion before t
 
 ### Goals
 
-* Introduce `NoExecute` taint for the `MachinePool` that triggers deletion of all bound `Machine`s.
+* Introduce `NoExecute` taint for the `MachinePool` that triggers deletion of all bound `Machine`s that do not tolerate the taint.
 
 ### Non-Goals
 
 * Handle the unplanned failure case: when a pool becomes unhealthy (IEP-15 marks it `Ready=Unknown`), automatically evict machines after a configurable threshold.
 * Drain orchestration (serial eviction, disruption budgets). Drain is a usage pattern on top: cordon (`NoSchedule`), then evict (`NoExecute`).
 * Automatic rescheduling of evicted machines. `machinePoolRef` is immutable: Recreating a Machine on another pool requires an external orchestrator which is out of scope.
-* Toleration-based selective eviction. `NoExecute` evicts all machines unconditionally.
 * Live migration of VMs between pools.
 * Resource pressure eviction.
 * Volume and NetworkInterface lifecycle on eviction. These follow existing ownership semantics and are not changed by this IEP.
@@ -68,7 +67,6 @@ In Kubernetes, when a `NoExecute` taint is added to a Node, the taint eviction c
 This IEP applies the same pattern to IronCore: a taint eviction controller deletes `Machine` objects when a `NoExecute` taint is set on a `MachinePool`. The poollet sees the `deletionTimestamp`, calls the provider to shut down the VM, and removes its finalizer to release the object.
 
 Key differences from Kubernetes:
-- No toleration matching: `NoExecute` evicts all machines unconditionally. In Kubernetes, a Pod that tolerates the `NoExecute` taint is retained (indefinitely, or for `tolerationSeconds` if set). IronCore does not honor tolerations for now.
 - Rescheduling is not built-in. Kubernetes has Deployments/ReplicaSets that recreate deleted Pods. IronCore does not yet provide an equivalent layer for recreating evicted Machines.
 - Kubernetes also has kubelet-initiated eviction: when a node runs low on resources (memory, disk, PIDs), the kubelet kills pods locally and marks them as `Failed` with reason `Evicted`. This is a separate mechanism from taint-based eviction. The kubelet acts autonomously without the control plane issuing a DELETE. Resource pressure eviction is explicitly out of scope for this IEP.
 
@@ -86,11 +84,11 @@ const (
 ```
 
 `NoSchedule` prevents new machines from being scheduled.
-`NoExecute` additionally signals that all machines on the pool must be deleted.
+`NoExecute` additionally signals that machines on the pool must be deleted unless they tolerate the taint.
 
 ### Eviction Flow
 
-A new **taint eviction controller** watches `MachinePool` taint changes. When a `NoExecute` taint is added, it issues DELETE on all `Machine`s bound to that pool.
+A new **taint eviction controller** watches `MachinePool` taint changes. When a `NoExecute` taint is added, it issues DELETE on every `Machine` bound to that pool that does not have a matching toleration. A `Machine` tolerates the taint if its `spec.tolerations` contains an entry matching the taint's `key`, `value`, and `effect`; tolerating machines are retained.
 
 The poollet already sets a finalizer on every Machine:
 
@@ -103,7 +101,7 @@ metadata:
 This finalizer ensures the object is not removed until the poollet confirms the VM is shut down.
 
 1. NoExecute taint added to MachinePool
-2. Taint eviction controller: DELETE Machine 
+2. Taint eviction controller: DELETE Machine if it has no matching toleration
 3. Machine gets deletionTimestamp (finalizer blocks removal)
 4. Poollet reconciles machine
    1. Calls provider and deletes machine
--- a/ieps/21-machine-eviction.md
+++ b/ieps/21-machine-eviction.md
@@ -1,5 +1,5 @@
 ---
-title: MachinePool Eviction
+title: Machine Eviction
 
 iep-number: 21
 
@@ -15,10 +15,11 @@ reviewers:
 
 - "@friegger"
 - "@gonzolino"
+- "@afritzler"
 
 ---
 
-# IEP-21: MachinePool Eviction
+# IEP-21: Machine Eviction
 
 ## Table of Contents
 
@@ -40,7 +41,7 @@ The provider handles graceful VM shutdown via its existing finalizer before the 
 
 ## Motivation
 
-Currently, when a bare-metal host backing a `MachinePool` needs maintenance, there is no way to signal this. The host gets shut down, VMs and workloads are killed, while the `Machine` objects in the API still appear as running.
+Currently, when a server backing a `MachinePool` needs maintenance, there is no way to signal this. The host gets shut down, VMs and workloads are killed, while the `Machine` objects in the API still appear as running.
 
 There is no mechanism to gracefully shut down VMs in an ordered fashion before the host goes down.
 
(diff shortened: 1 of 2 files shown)


ieps/21-machinepool-eviction.md
```diff
@@ -0,0 +1,122 @@
...
+- Kubernetes also has kubelet-initiated eviction: when a node runs low on resources (memory, disk, PIDs), the kubelet kills pods locally and marks them as `Failed` with reason `Evicted`. This is a separate mechanism from taint-based eviction. The kubelet acts autonomously without the control plane issuing a DELETE. Resource pressure eviction is explicitly out of scope for this IEP.
+
+ref: [kubernetes taint eviction](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/tainteviction/taint_eviction.go#L147)
+### API Changes
+
+Extend the `TaintEffect` type in `api/common/v1alpha1/common_types.go`:
+
+```go
+const (
+    TaintEffectNoSchedule TaintEffect = "NoSchedule"
```
friegger: Probably this should be added only once it is used.

ieps/21-machinepool-eviction.md
```diff
@@ -0,0 +1,122 @@
...
+  - machinepoollet.ironcore.dev/machine
+```
+
+This finalizer ensures the object is not removed until the poollet confirms the VM is shut down.
+
+1. NoExecute taint added to MachinePool
+2. Taint eviction controller: DELETE Machine 
+3. Machine gets deletionTimestamp (finalizer blocks removal)
+4. Poollet reconciles machine
+   1. calls provider and deletes machine
```
friegger:
```suggestion
   1. Calls provider and deletes machine
```

ieps/21-machine-eviction.md
```diff
@@ -0,0 +1,122 @@
...
+
+This IEP applies the same pattern to IronCore: a taint eviction controller deletes `Machine` objects when a `NoExecute` taint is set on a `MachinePool`. The poollet sees the `deletionTimestamp`, calls the provider to shut down the VM, and removes its finalizer to release the object.
+
+Key differences from Kubernetes:
+- No toleration matching: `NoExecute` evicts all machines unconditionally.
+- Rescheduling is not built-in. Kubernetes has Deployments/ReplicaSets that recreate deleted Pods. IronCore does not yet have an equivalent higher-level controller.
+- Kubernetes also has kubelet-initiated eviction: when a node runs low on resources (memory, disk, PIDs), the kubelet kills pods locally and marks them as `Failed` with reason `Evicted`. This is a separate mechanism from taint-based eviction. The kubelet acts autonomously without the control plane issuing a DELETE. Resource pressure eviction is explicitly out of scope for this IEP.
+
+ref: [kubernetes taint eviction](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/tainteviction/taint_eviction.go#L147)
```
friegger, line 73:
```suggestion
ref: [kubernetes taint eviction](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/tainteviction/taint_eviction.go#L147)

```

ieps/21-machinepool-eviction.md
lukasfrank: We already have this

ieps/21-machine-eviction.md
```diff
@@ -0,0 +1,123 @@
...
+  - machinepoollet.ironcore.dev/machine
+```
+
+This finalizer ensures the object is not removed until the poollet confirms the VM is shut down.
+
+1. NoExecute taint added to MachinePool
+2. Taint eviction controller: DELETE Machine 
+3. Machine gets deletionTimestamp (finalizer blocks removal)
+4. Poollet reconciles machine
+   1. Calls provider and deletes machine
```
gonzolino, line 107: Do we need to care for the case that the provider fails to delete the machine?

ieps/21-machine-eviction.md
```diff
@@ -0,0 +1,123 @@
...
+    TaintEffectNoExecute  TaintEffect = "NoExecute"
+)
+```
+
+`NoSchedule` prevents new machines from being scheduled.
+`NoExecute` additionally signals that all machines on the pool must be deleted.
+
+### Eviction Flow
+
+A new **taint eviction controller** watches `MachinePool` taint changes. When a `NoExecute` taint is added, it issues DELETE on all `Machine`s bound to that pool.
```
gonzolino: Do we need some kind of mechanism to keep track of the machine deletion? So that an operator knows if the eviction is complete, or what the status is (e.g. 3/10 machines evicted).

ieps/21-machine-eviction.md
```diff
@@ -0,0 +1,123 @@
...
+3. Machine gets deletionTimestamp (finalizer blocks removal)
+4. Poollet reconciles machine
+   1. Calls provider and deletes machine
+   2. Removes finalizer
+5. API server removes object
+
+
+## Alternatives
+
+### Keep Machine with an `Evicted` state instead of deleting
```
gonzolino, line 114: I am still a fan of this alternative. I understand there is additional complexity, but I think it is confusing for users if their machines just magically disappear.

I know k8s just deletes Pods, but I would argue there is a difference in the user expectation towards the lifecycle of Pods and VMs....

ieps/21-machine-eviction.md
lukasfrank: We already know it through the `machinePoolRef` and the deletion timestamp.  Additionally we can emit an event.

ieps/21-machine-eviction.md
gonzolino: Ok, after thinking about this a bit more and having some discussions I am no longer a fan of this alternative 😅 

Marking the terminal state can only be done by the poollet. But for the poollet to take action, the taint eviction controller needs to mark Machine objects for eviction first. Since the controller can't delete the Machine object (since we want to keep it in terminal state), we need to adapt the API for it. So now draining machines from a machine pool requires special handling between taint eviction controller and poollet.

The original proposal is a lot cleaner: taint eviction controller deletes the Machine object and the poollet invokes the provider to do a regular machine deletion (same workflow as if a user or some other tooling deletes a Machine object).

ieps/21-machine-eviction.md
lukasfrank: imo the eviction controller's responsibility ends at deleting the `Machine`. The poolet already tries continuously to delete the `Machine` and if it is stuck in terminating, an operator might need to resolve the issue. wdyt?

ieps/21-machine-eviction.md
gonzolino: Yes on the Machine objects we can see it. But my question was more targeted towards the overall eviction progress. Do we need to keep track when all machines are actually gone? Or does the eviction controller simply issue DELETEs and then doesn't care if the eviction actually finishes?

lukasfrank: ah okay, got it. Same as Kubernetes, the taint eviction controller issues DELETEs and is done. It doesn't track completion. The controller which initiated the eviction needs to track it.

gonzolino (approved):
lgtm

ieps/21-machinepool-eviction.md
```diff
@@ -0,0 +1,123 @@
...
+## Summary
+
+This IEP defines how `Machines` are evicted from a MachinePool.
+
+We introduce a `NoExecute` taint effect. When set on a `MachinePool`, a taint eviction controller issues DELETE on all bound `Machine`s. 
+The provider handles graceful VM shutdown via its existing finalizer before the object is fully removed.
+
+## Motivation
+
+Currently, when a bare-metal host backing a `MachinePool` needs maintenance, there is no way to signal this. The host gets shut down, VMs and workloads are killed, while the `Machine` objects in the API still appear as running.
```
afritzler: I would drop here the bare-metal and just use host/server.

ieps/21-machine-eviction.md
```diff
@@ -0,0 +1,123 @@
...
+This IEP defines how `Machines` are evicted from a MachinePool.
+
+We introduce a `NoExecute` taint effect. When set on a `MachinePool`, a taint eviction controller issues DELETE on all bound `Machine`s. 
+The provider handles graceful VM shutdown via its existing finalizer before the object is fully removed.
+
+## Motivation
+
+Currently, when a bare-metal host backing a `MachinePool` needs maintenance, there is no way to signal this. The host gets shut down, VMs and workloads are killed, while the `Machine` objects in the API still appear as running.
+
+There is no mechanism to gracefully shut down VMs in an ordered fashion before the host goes down.
```
afritzler, line 46: Isn't the eviction process as proposed here more like a "random remove all machines"?

ieps/21-machinepool-eviction.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Goals
+
+* Introduce `NoExecute` taint for the `MachinePool` that triggers deletion of all bound `Machine`s.
+
+### Non-Goals
+
+* Handle the unplanned failure case: when a pool becomes unhealthy (IEP-15 marks it `Ready=Unknown`), automatically evict machines after a configurable threshold.
+* Drain orchestration (serial eviction, disruption budgets). Drain is a usage pattern on top: cordon (`NoSchedule`), then evict (`NoExecute`).
+* Automatic rescheduling of evicted machines. `machinePoolRef` is immutable: Recreating a Machine on another pool requires a higher-level controller (or Gardener) which is out of scope.
```
afritzler: Instead of high-level controller or Gardener I would say something like "external orchestrator".

ieps/21-machinepool-eviction.md
```diff
@@ -0,0 +1,123 @@
+---
+title: MachinePool Eviction
```
afritzler: Maybe we should call this proposal: Machine Eviction as we describe how to evict machines?

ieps/21-machinepool-eviction.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Comparison to Kubernetes
+
+In Kubernetes, when a `NoExecute` taint is added to a Node, the taint eviction controller (part of the NodeLifecycleController) deletes all Pods that do not tolerate the taint. The kubelet sees the `deletionTimestamp` on each Pod, sends SIGTERM to containers, waits for the grace period, then reports back. The Pod object is removed once the kubelet confirms termination.
+
+This IEP applies the same pattern to IronCore: a taint eviction controller deletes `Machine` objects when a `NoExecute` taint is set on a `MachinePool`. The poollet sees the `deletionTimestamp`, calls the provider to shut down the VM, and removes its finalizer to release the object.
+
+Key differences from Kubernetes:
+- No toleration matching: `NoExecute` evicts all machines unconditionally.
+- Rescheduling is not built-in. Kubernetes has Deployments/ReplicaSets that recreate deleted Pods. IronCore does not yet have an equivalent higher-level controller.
```
afritzler: "IronCore does not yet provide an equivalent layer for recreating evicted Machines" instead of high level controller.

afritzler (changes requested):
Thanks @lukasfrank I added my comments to the document.

One thing maybe to rephrase is the fact that `NoExecute` evicts all bound Machines unconditionally and does not honor tolerations, unlike Kubernetes where `NoExecute`-tolerating Pods are retained (with TolerationSeconds). This is a deliberate simplification. Users should be aware that a `NoExecute` toleration on a Machine has no effect. If selective retention is needed later, it would require the eviction controller to consult tolerations.

ieps/21-machine-eviction.md
lukasfrank: It deletes all `Machine`s assigned to the `Pool` where the pool provider ensures that the graceful shutdown and deletion is done.

ieps/21-machine-eviction.md
```diff
@@ -0,0 +1,124 @@
...
+## Proposal
+
+### Comparison to Kubernetes
+
+In Kubernetes, when a `NoExecute` taint is added to a Node, the taint eviction controller (part of the NodeLifecycleController) deletes all Pods that do not tolerate the taint. The kubelet sees the `deletionTimestamp` on each Pod, sends SIGTERM to containers, waits for the grace period, then reports back. The Pod object is removed once the kubelet confirms termination.
+
+This IEP applies the same pattern to IronCore: a taint eviction controller deletes `Machine` objects when a `NoExecute` taint is set on a `MachinePool`. The poollet sees the `deletionTimestamp`, calls the provider to shut down the VM, and removes its finalizer to release the object.
+
+Key differences from Kubernetes:
+- No toleration matching: `NoExecute` evicts all machines unconditionally. In Kubernetes, a Pod that tolerates the `NoExecute` taint is retained (indefinitely, or for `tolerationSeconds` if set). IronCore does not honor tolerations for now.
```
adracus: Why do we not want to honor the tolerations? For the taint eviction controller, this should be a simple filter step, shouldn't it?

adracus (commented):
I really like this proposal! If you could check with this tolerations comment that'd be perfect 😊

ieps/21-machine-eviction.md
lukasfrank: You're right that it's technically just a filter step, so it's cheap to add. The reason I left it out of scope is rather than difficulty:
-  the reason for eviction right now is "get everything off this pool" so it can be maintained. If we honored a toleration, what should then happen with the `Machine` that stays on a host we're trying to drain for maintenance? 
-  deferring is forward-compatible and additive, so when we need it we can add it

adracus (approved):
🚀

ieps/21-machine-eviction.md
lukasfrank: We decided to implement tolerations in an offline discussion since the `Machine` already have `Tolerations []commonv1alpha1.Toleration` in the spec

afritzler (approved):

PR #42, opened 2026-06-30
Add IEP for Network Fabric Automation

hardikdr:


ieps/tbd-network-fabric-automation.md
```diff
@@ -0,0 +1,469 @@
...
+## Motivation
+
+Switches in a data center have a lifecycle. They get racked, they get configured,
+they run for years, and eventually they get replaced or repurposed. Today, managing
+this lifecycle involves a mix of ZTP templates, manual YAML, and tribal knowledge
+about which switches are available, which are in use, and which are in maintenance.
+
+The goal is to bring this under Kubernetes-native control: declarative, observable,
+and automatable.
+
```
oberlademeister, line 69: These observations are too general to be clearly wrong. However, they also do not explain what the author tries to accomplish here. I would want to see stronger motivation with a clear ironcore related outcome in mind.

ieps/tbd-network-fabric-automation.md
```diff
@@ -0,0 +1,469 @@
...
+- **Vendor-agnostic by design.** The core API is neutral. Vendor differences live
+  behind a provider interface, not in the API itself.
+
+- **Automate configuration across a fleet.** A single intent expressed at the fabric
+  level should be able to configure dozens of switches consistently, not one YAML at
+  a time.
+
+- **Start small and grow.** The initial object model is the minimum needed to be
+  useful. New capabilities are added when real use cases justify them.
+
```
oberlademeister, line 84: These are great general goals. What would be the small start exactly? How would we grow?

ieps/tbd-network-fabric-automation.md
```diff
@@ -0,0 +1,469 @@
...
+- **Modeling a specific topology.** CLOS, spine-leaf, flat L2 — the API should not
+  be hardwired to one physical arrangement. Topology assumptions can live in higher-
+  level controllers (e.g., Fabric), not the core objects.
+
+- **Exposing every knob.** The goal is not a Kubernetes projection of gNMI or NETCONF.
+  We are not building a general-purpose switch configuration tool.
+
+- **Finalizing the API here.** This document is a concept. The actual field names,
+  resource shapes, and versioning will be decided when we implement.
+
```
oberlademeister, line 100: I agree to "not expose every knob". I would actually not want to expose "any knob at all" because any knob is vendor specific.

ieps/tbd-network-fabric-automation.md
```diff
@@ -0,0 +1,469 @@
...
+  prefixes:
+  - 10.0.0.0/64
+  vlans:
+  - id: 1000
+    prefix: 10.0.100.0/24
+    ..  
+  bgp:
+    asn: 0001
+    peerGroups:
+      - name: leafs
```
oberlademeister, line 230: How is that not vendor-specific and design-specific? What would be the desired behaviour if you change any of this data?

ieps/tbd-network-fabric-automation.md
```diff
@@ -0,0 +1,469 @@
...
+  - type: ConfigApplied
+    status: "True"
+    observedGeneration: 4
+```
+
+To reconfigure a switch, change VLAN ranges, update BGP ASN, update the `Switch` spec.
+The controller detects the generation change and pushes the delta to the device.
+To rebind a switch to different hardware, delete and recreate the `Switch` object.
+This keeps reconfiguration explicit and auditable.
+
```
oberlademeister, line 255: Is this for a lab setting? In any realistic scenario the design of a (pre)prod scenario is going to be static. I do not understand the scenario.

ieps/tbd-network-fabric-automation.md
```diff
@@ -0,0 +1,469 @@
...
+For stable, well-understood topologies: VLAN-per-port, BGP unnumbered spine-leaf,
+config changes after bootstrap are rare by design. In these environments, any
+modification to a running switch should be deliberate, reviewed, and applied through
+a controlled path. For these topologies, the `Fabric` object carries an
+`immutabilityPolicy: Strict` that makes this explicit: the Fabric controller owns
+the `Switch` objects it creates, and any direct modification is reconciled back
+immediately. The `Switch` objects become the rendered output of the Fabric intent,
+not something an operator edits by hand.
+
```
oberlademeister, line 354: In IronCore, do we no what exactly is fixed and what will change over time?

ieps/tbd-network-fabric-automation.md
```diff
@@ -0,0 +1,469 @@
...
+---
+
+## Alternatives
+
+**General-purpose gNMI projection.**
+Exposing every switch knob as a Kubernetes resource makes the API surface
+unbounded and tightly couples the Kubernetes API to vendor-specific data models.
+We want automation for specific, well-understood use cases not a Kubernetes
+wrapper around the entire gNMI tree.
```
oberlademeister, line 471: Again. Too general to be reviewed. This can be anything to anyone. Should we not rather take a use-case based approach that talking about this so general that it is - to me - no longer meaningful.

oberlademeister (changes requested):
As with the individual comments, I have issues with the generality of this approach without providing at least step one of what would be implemented for one example vendor. In the IETF, it is a good principle for protocol design that for anything aiming to be vendor or implementation specific, at least two real implementations need to exist and interoperate. I am questioning the approach if it is not clear we would do this for two vendors.

ieps/tbd-network-fabric-automation.md
hardikdr: Let me break it down here. 
- Declarative: We would like to manage the lifecycle of the Switches using the CRDs in Kubernetes, with a declarative intent. This primarily means onboarding the switch in an automated fashion, allowing the possibility to reconfigure where it makes sense, and potentially day-2 operations. 
- Observable: With the point above, with Kubernetes primitives, we can observe the state better without needing to have imperative tooling. 
- Automatable: When you have the ones above, higher-level automations become more natural, see the Fabric object that would allow automating larger numbers of devices. 

To avoid the risk of over-expanding the proposal, I would keep the statement above precise. 
As Ironcore is Kubernetes, inside out, and this naturally becomes the positive outcome for Ironcore that such network devices are under the realm of Kubernetes.

ieps/tbd-network-fabric-automation.md
hardikdr: Let me offer a precise answer here. 
Starting small, and growing means, we start with the small number of Kuberentes CRDs, and organically grow them as usecases drive. 
Concrete example: We start with only `Device`, `Switch`, `SwitchInterface`, and introduce API which were proposed in the original wire-proposal such as `Server`, `Host`, `ServerInterface`, and `DeviceInterface` later as usecases drive them.

Ref: https://github.com/ironcore-dev/enhancements/pull/42/changes#diff-0399095f1681ab10ceb9be8ab14433994e7443d2d56dd98b9a4ee14226ae8877R115

ieps/tbd-network-fabric-automation.md
hardikdr: Conceptually, this is inspired by the wire-proposal which was prepared under your guidance, I believe, and which you have admired. See: https://github.com/ironcore-dev/enhancements/pull/21/changes#diff-2823c50794ea2ff155a1d51203d65403465c89e7daca3e7f5e7c8cb29c335602R318

“ not exposing any knobs" is a very broad statement; could you please share more context and evidence/reasoning behind why you think so? 
In my opinion, a hybrid model, where it makes sense, we should consider bringing it to the API surface. I am not too strict about bgp etc though.

ieps/tbd-network-fabric-automation.md
hardikdr: Again, https://github.com/ironcore-dev/enhancements/pull/21/changes#diff-2823c50794ea2ff155a1d51203d65403465c89e7daca3e7f5e7c8cb29c335602R318, same answer as above.

ieps/tbd-network-fabric-automation.md
hardikdr: I really liked the perspective here in this comment on the original wire proposal: https://github.com/ironcore-dev/enhancements/pull/21/changes#r3378707287

In an ideal world, with the kind of architecture we have, I would love to see Switches being static, and the best part is, "Kubernetes doesnt modify the CRD by itself unless we modify it". Prod being static should be the "process goal" and should not be baked into the API/software itself, that'd be a major anti-pattern.

ieps/tbd-network-fabric-automation.md
hardikdr: Not only Ironcore, but in general for any automation, we wouldn't know exactly what can potentially change, and that's precisely the reason the proposal is about an extensible, vendor-agnostic design which can incorporate future requirements without having to throw away the existing software.

ieps/tbd-network-fabric-automation.md
hardikdr: Please feel free to point out what is too general, and we can discuss. The idea is simple, we dont want the entire gNMI / Yang model to pop up in the Kubernetes surface, that's an overkill, it should be deliberate, well-understood aspects that come to the API surface that are realistic to automate.

ieps/tbd-network-fabric-automation.md
```diff
@@ -0,0 +1,471 @@
...
+
+iep-number: tbd
+
+creation-date: 2026-06-29
+
+status: draft
+
+authors:
+
+- "@hardikdr"
```
hardikdr, line 12: @adracus @afritzler @Giluerre As mentioned in the last weekly sync, please feel free to add yourself to the authors here if you feel resonated to the extended proposal.

hardikdr:
Closing this in favor of the original wire-proposal. We will try to incorporate new aspects there.

PR #49, opened 2026-07-20
IronCore Image V2 proposal

adracus:
# Proposal

Our current image specification has evolved a lot since the beginning.
This proposal tries to cleanly describe how we boot and to enhance upon the ideas we have so far.

### DIFF
--- a/ieps/tbd-ironcore-image-v2.md
+++ b/ieps/tbd-ironcore-image-v2.md
@@ -78,9 +78,11 @@ functionality using e.g. initramfs layering.
   config media and manifest type:
   * UKI-like boot mode
   * Disk boot mode
+  * EFI executable boot mode
 * Don't specify UKIs as UKI binary blobs anymore. Instead,
   specify their components (kernel, initrd, stub etc.) as distinct layers,
   allowing for layer deduplication.
+* Still allow passing in EFI executables directly for non-UKI use cases.
 * Create facilities for installer images: Namely, composability of initramfs
   and simple reboot to disk after finishing disk writing.
 
@@ -150,8 +152,7 @@ being appended via default media type extension (e.g. `+xz`).
 
 ### EFI executable booting
 
-EFI executable booting just allows specifying a raw EFI
-executable to boot.
+EFI executable booting allows specifying an EFI executable to boot.
 
 #### Manifest
 
@@ -170,6 +171,14 @@ executable to boot.
 ```json
 {}
 ```
+
+#### EFI Executable
+
+The EFI executable, required.
+
+**Media Type**
+
+`application/vnd.ironcore.efi.executable`
 
 ### Disk booting
 
--- a/ieps/tbd-ironcore-image-v2.md
+++ b/ieps/tbd-ironcore-image-v2.md
@@ -51,8 +51,6 @@ different ways to boot:
 * [Direct kernel + initramfs](https://github.com/ironcore-dev/ironcore-image/blob/f89db68d5e9323ed5d6024ba70897e077504c25b/image.go#L19)
 * [Squashfs](https://github.com/ironcore-dev/ironcore-image/blob/main/image.go#L20) 
 
-There's even a 'missing' way to boot: 'Raw' EFI executables.
-
 ### Lack of Layer Deduplication
 
 For UKI booting, since we allow specifying the entire UKI, we cannot make use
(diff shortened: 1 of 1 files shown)


ieps/tbd-ironcore-image-v2.md
```diff
@@ -0,0 +1,214 @@
...
+
+#### EFI Executable
+
+The EFI executable, required.
+
+**Media Type**
+
+`application/vnd.ironcore.efi.executable`
+
+### Disk booting
```
maxmoehl, line 183: I would prefer if we could keep supporting raw in addition to qcow2.

* It is the format that all our current images are built in and breaking with it requires us to either support both old and new images anyway or rebuild all images.
* The de-facto storage backend of IronCore is ceph which requires a raw file anyway.
* It's an incredibly simple format and battle-tested with IronCore, we know it works reliably.

Note that I'm not arguing against qcow2, if anyone needs the more advanced feature set, they should pick that.

ieps/tbd-ironcore-image-v2.md
```diff
@@ -0,0 +1,214 @@
...
+
+Within our single manifest and configuration type, we house multiple
+different ways to boot:
+
+* [UKI](https://github.com/ironcore-dev/ironcore-image/blob/f89db68d5e9323ed5d6024ba70897e077504c25b/image.go#L21)
+* [ISO](https://github.com/ironcore-dev/ironcore-image/blob/f89db68d5e9323ed5d6024ba70897e077504c25b/image.go#L22)
+* [Direct kernel + initramfs](https://github.com/ironcore-dev/ironcore-image/blob/f89db68d5e9323ed5d6024ba70897e077504c25b/image.go#L19)
+* [Squashfs](https://github.com/ironcore-dev/ironcore-image/blob/main/image.go#L20) 
+
+There's even a 'missing' way to boot: 'Raw' EFI executables.
```
maxmoehl: I'd drop this, the current UKI layer is for all intents and purposes booting any EFI binary, it's just a mislabeling on our end.

ieps/tbd-ironcore-image-v2.md
adracus: I'd not extend the API surface with a layer that is equal to the QCOW layer if we're not using it. Towards your points:
* This is a major change. We anyways have to build new images. And yes, we will have a transition period where we support both old and new.
* Ceph is the storage backend of IaaS. With the new proposal, disk boot will now also be possible for metal, so metal is a consumer that has to be handled as well. Ceph does not require a raw file but a raw source. Qcow allows for that.
* RAW allocates the full space for sparse files, does not have any support for layering which allows for content dedup which drastically improves dev experience, storage pressure on registries and potentially also download speed (depending on e.g. local pull-through caches). Why should qcow work any less reliably as well?

Also, any OS distro out there already has qcow, which makes it extremely simple to use what's out there already. Additionally, converting raw to qcow is as simple as

```bash
qemu-img convert -f raw -O qcow2 -p src.raw dst.qcow2
```

So, all in all, there is no benefit from keeping raw as the inferior format around.

ieps/tbd-ironcore-image-v2.md
adracus: I mean originally the idea to also support efi executables came from you, so I'll just drop it.
However, if somebody wants to boot a non-UKI EFI-executable, this is not possible anymore after dropping it, just clarifying it here.

maxmoehl: What? The comment is just about that one sentence. We should keep the raw EFI feature of v2.

adracus: Oh then I got that wrong 🙈  sorry for that - fixed it!

lukasfrank (commented):
Isn't one essential point of an UKI that it can be signed and shipped as one unit (including the cmdline). Can you elaborate on how this should work here if we split up the units.

ieps/tbd-ironcore-image-v2.md
```diff
@@ -0,0 +1,211 @@
...
+While working fine for our implementation so far, there are various important
+shortcomings:
+
+### Boot Modes
+
+Within our single manifest and configuration type, we house multiple
+different ways to boot:
+
+* [UKI](https://github.com/ironcore-dev/ironcore-image/blob/f89db68d5e9323ed5d6024ba70897e077504c25b/image.go#L21)
+* [ISO](https://github.com/ironcore-dev/ironcore-image/blob/f89db68d5e9323ed5d6024ba70897e077504c25b/image.go#L22)
```
lukasfrank: ISO boot silently disappears, what happens to ISO boot?

ieps/tbd-ironcore-image-v2.md
lukasfrank: While reading the proposal, it also came to my mind: Why not keeping the raw format?

ieps/tbd-ironcore-image-v2.md
adracus: ISO as source won't be supported anymore. However, ISO as a 'transport mechanism' will become more relevant in the metal-operator.

PR #50, opened 2026-07-21
New proposal for dpservice part of underlay-encryption

mast-wch:
This is a new proposal for the dpservice part of the [underlay-encryption proposal](https://github.com/ironcore-dev/enhancements/pull/38) as discussed in the network community meetings. It focuses on the encryption of dpservice instances and does not handle the key exchange any more.

### DIFF
--- a/ieps/NNNN-underlay-encryption-dpservice.md
+++ b/ieps/NNNN-underlay-encryption-dpservice.md
@@ -1,0 +1,182 @@
+---
+title: Underlay Encryption - IPsec encryption between compute hosts
+
+iep-number: NNNN
+
+creation-date: 2026-07-15
+
+status: implementable
+
+authors:
+
+- "@kitsudaiki"
+- "@mast-wch"
+- "@markus-hentsch"
+
+reviewers:
+
+- "tbd"
+
+---
+
+# IEP-NNNN: Underlay Encryption - IPsec encryption between compute hosts
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Alternatives](#alternatives)
+
+## Summary
+
+Implementation of an underlay encryption for the traffic between compute hosts. 
+This touches on the networking component dpservice in the first place. In further iterations it also touches instances that use the Linux kernel for their networking.
+
+## Motivation
+
+Currently, the underlay network does not provide encryption. 
+However, to align with regulatory requirements for specific security domains (for example as mandated by the BSI), encryption is necessary. 
+Once enabled, all underlay traffic originating from dpservice instances must be encrypted to ensure data confidentiality and integrity.
+
+### Goals
+
+- Add the possibility to encrypt IPv6-encapsulated traffic between dpservice instances using IPsec ESP in transport mode
+
+### Non-Goals
+
+- This IEP does not handle the topic of key exchange. It is assumed that a symmetric key and salt will be made available to both 
+  communication ends of a Security Association. 
+
+- The decision of whether to encrypt outgoing packets in dpservice will rely on a sound decision in the control plane. That means that when no encryption for a combination of underly IPv6 address and VNI is configured packets will be sent unencrypted.
+
+- The option to offload encryption to the network card will be part of a subsequent enhancement.
+  Melanox BlueField network cards also support crypto offloading for IPsec (<https://docs.nvidia.com/doca/sdk/ipsec-crypto-offload/index.html>). 
+  This will require a secondary version of the implementation in the code, because the workflow between the CPU and the offloaded version is different. 
+
+- This IEP aims for East-West traffic encryption between compute hosts running dpservice. North-South traffic between compute hosts and routers will not be encrypted and thus routers will not be in the focus. However, all changes made here shall avoid complications with future integration with the Linux kernel XFRM interface.
+
+## Proposal
+
+### Related open issues
+
+There are already 2 open issues related to this topic:
+
+- https://github.com/ironcore-dev/metalnet/issues/320
+- https://github.com/ironcore-dev/roadmap/issues/69
+
+This IEP was a first draft that combined key exchange and encryption in one draft:
+- https://github.com/ironcore-dev/enhancements/pull/38
+
+### Overview:
+
+```mermaid
+---
+config:
+---
+graph TD
+    router["Router"]
+    control_plane1["Control Plane"]
+    control_plane2["Control Plane"]
+
+    dpservice01["Dpservice"]
+    dpservice02["Dpservice"]
+
+    
+
+    router <---> |IPinIPv6| dpservice01
+    router <---> |IPinIPv6| dpservice02
+
+    dpservice01 <--> |IPsec| dpservice02 
+
+    subgraph Compute Host 01
+        control_plane1 --> |gRPC| dpservice01
+    end
+
+    subgraph Compute Host 02
+        control_plane2 --> |gRPC| dpservice02
+    end
+
+```
+
+### dpservice
+
+The DPDK graph in dpservice has to be extended with encryption and decryption nodes. 
+Those nodes shall use the IPSec implementation provided by [DPDK](https://doc.dpdk.org/guides/prog_guide/ipsec_lib.html). 
+The library already provides the means for AES-256-GCM encryption/decryption as well as a SA database. 
+It also handles sequence numbers to protect against replay attacks. 
+The DPDK library will be configured to use ESP in transport mode for the single SAs. 
+As there is already encapsulation and decapsulation in place, this will basically result in tunnel mode ESP.
+
+Packets will be encrypted on the sender side after the IPv6-encapsulation and before the decapsulation on the receiver side. 
+This ensures, that only traffic leaving/entering the physical host, will be encrypted and decrypted to avoid unnecessary packet processing overhead. 
+An SA is created for each remote compute host, or rather remote dpservice instance, with which the local dpservice has to exchange packets. 
+These SAs are stored in the DPDK SA database. 
+The key as well as the salt (see [RFC4106](https://datatracker.ietf.org/doc/html/rfc4106)) for the encrypted connection is provided over the gRPC connection. 
+Whenever a new key is pushed, a new SA gets created. Pushing a new key for an established SA will be used for key rotation.
+
+For backwards compatibility the encryption must be configurable for each pair of compute hosts. That is, as long as a route to another compute host is configured without an SA, the traffic will further be unencrypted until encryption is enabled.
+
+#### Scope of Encryption
+As of now dpservice already sends IP-in-IPv6 packets in the underlay network. This shall not be changed at the moment. Rather we intend to encrypt that encapsulated package as a whole. Following [RFC4303](https://www.rfc-editor.org/info/rfc4303/) (ESP in transport mode), the resulting package structure would be this:
+
+```
+| IPv6 Header | ESP Header | Inner IPv4 Package | ESP Trailer | ICV* |
+<-------Unencrypted------->                                   <Unenc.>
+               <---------------Authenticated------------------>
+                            <-------------Encrypted----------->
+
+*ICV - Integrity Check Value 
+```
+
+#### Key rotation
+During key rotation there will need to be two SAs in place between a pair of compute hosts. The old one that shall be superseded and the new one.
+The new one needs to be active on the receiver side before the sender can start sending with it. The old one needs to be in place on the receiver side for a grace period beginning with the moment when it is deleted from the sender side for as long as packets may still be on the wire. Coordinating this is the responsibility of the control plane that handles the key management.
+
+#### Hardware support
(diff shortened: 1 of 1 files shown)


ieps/NNNN-underlay-encryption-dpservice.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Summary
+
+Implementation of an underlay encryption for the traffic between compute hosts. 
+This touches on the networking component dpservice in the first place. In further iterations it also touches instances that use the Linux kernel for their networking.
+
+## Motivation
+
+Currently, the underlay network does not provide encryption. 
+However, to align with regulatory requirements for specific security domains (for example as mandated by the BSI), encryption is necessary.
```
maxmoehl: The motivation is pretty thin and hinges on this one sentence, please provide a source when you reference that external parties mandate something.

I've also noticed that an alternative listed is to only encrypt the payload of the overlay (e.g. not the overlay headers themselves). If that is an alternative, I'm wondering why you don't just rely on application level encryption? Most traffic these days is anyway encrypted via (m)TLS so this seems like a much simpler option then.

ieps/NNNN-underlay-encryption-dpservice.md
```diff
@@ -0,0 +1,182 @@
...
+
+### Non-Goals
+
+- This IEP does not handle the topic of key exchange. It is assumed that a symmetric key and salt will be made available to both 
+  communication ends of a Security Association. 
+
+- The decision of whether to encrypt outgoing packets in dpservice will rely on a sound decision in the control plane. That means that when no encryption for a combination of underly IPv6 address and VNI is configured packets will be sent unencrypted.
+
+- The option to offload encryption to the network card will be part of a subsequent enhancement.
+  Melanox BlueField network cards also support crypto offloading for IPsec (<https://docs.nvidia.com/doca/sdk/ipsec-crypto-offload/index.html>).
```
maxmoehl:
```suggestion
- The option to offload encryption to the network card will be part of a subsequent enhancement.
```

We are not tied to any specific hardware (though we do only test on NVIDIA NICs right now).

ieps/NNNN-underlay-encryption-dpservice.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Motivation
+
+Currently, the underlay network does not provide encryption. 
+However, to align with regulatory requirements for specific security domains (for example as mandated by the BSI), encryption is necessary. 
+Once enabled, all underlay traffic originating from dpservice instances must be encrypted to ensure data confidentiality and integrity.
+
+### Goals
+
+- Add the possibility to encrypt IPv6-encapsulated traffic between dpservice instances using IPsec ESP in transport mode
```
maxmoehl: When introducing abbreviations that have technical relevance please spell them out once.

```suggestion
- Add the possibility to encrypt IPv6-encapsulated traffic between dpservice instances using IPsec Encapsulating Security Payload (ESP) in transport mode
```

ieps/NNNN-underlay-encryption-dpservice.md
```diff
@@ -0,0 +1,182 @@
...
+The DPDK graph in dpservice has to be extended with encryption and decryption nodes. 
+Those nodes shall use the IPSec implementation provided by [DPDK](https://doc.dpdk.org/guides/prog_guide/ipsec_lib.html). 
+The library already provides the means for AES-256-GCM encryption/decryption as well as a SA database. 
+It also handles sequence numbers to protect against replay attacks. 
+The DPDK library will be configured to use ESP in transport mode for the single SAs. 
+As there is already encapsulation and decapsulation in place, this will basically result in tunnel mode ESP.
+
+Packets will be encrypted on the sender side after the IPv6-encapsulation and before the decapsulation on the receiver side. 
+This ensures, that only traffic leaving/entering the physical host, will be encrypted and decrypted to avoid unnecessary packet processing overhead. 
+An SA is created for each remote compute host, or rather remote dpservice instance, with which the local dpservice has to exchange packets.
```
maxmoehl: Does this still hold true with the discussion we recently have where you explained that we will have one key per VNI?

ieps/NNNN-underlay-encryption-dpservice.md
```diff
@@ -0,0 +1,182 @@
...
+    - Salt
+  - Traffic selector:
+    - VNI
+    - Source or Destination IPv6 Address (depends on Direction)
+- Is used to create a new Security Association between two endpoints
+
+Delete Security Association
+- Needs:
+  - SPI
+- Is used to remove a Security Association
```
maxmoehl: I would prefer if you could propose the necessary protobuf definitions with appropriate comments instead of prose.

maxmoehl (changes requested):

ieps/NNNN-underlay-encryption-dpservice.md
mast-wch: So noted on the first part - will add a viable reference once I have one.

Regarding the second part: The alternative is a reaction to the request from last weeks community call (see notes from 2026-07-15) where we got the question on whether it'd be possible to have visible inner IPv4 headers for possible maintenance and testing reasons. I'm hoping to get more input on that alternative and on whether it matches what was asked. I am with you that this would beat the purpose of the underlay encryption to a degree. However, it would still encrypt the (potentially not TLS encrypted) overlay traffic except for the headers.

ieps/NNNN-underlay-encryption-dpservice.md
mast-wch: The SAs are unidirectional and need to be established with each communication partner and on both ends. One outgoing (A -> B) and one incoming (A <- B) and vice versa on B's side. 
The negotiation of the key is moved to the other IEP regarding key exchange. To keep the dpservice part independent from that, it simply expects a key and a salt. Whether the key is reused for the whole VNI is not decided here, I'd say.

trevex: Even with group keys we have to derive individual SAs as per default replay protection is handled by a sequence number (same with Linux xfrm). This is the only replay protection that is offloadable in modern NICs.

Hopefully we get some modern time-based options in the future which would alleviate this...

ieps/NNNN-underlay-encryption-dpservice.md
christianhuening: Encrypting on the application layer is always an option. But it's precisely only an option and thus has to be enforced on an organisational level. I also know of applications where that would be difficult to implement and service meshes don't always work either or again have to be implemented enforced by a governance rule.

So long story short, having underlay encryption transparent to the cloud user (transparent link encryption or TLE) is highly desirable.

byteocean: > So noted on the first part - will add a viable reference once I have one.
> 
> Regarding the second part: The alternative is a reaction to the request from last weeks community call (see notes from 2026-07-15) where we got the question on whether it'd be possible to have visible inner IPv4 headers for possible maintenance and testing reasons. I'm hoping to get more input on that alternative and on whether it matches what was asked. I am with you that this would beat the purpose of the underlay encryption to a degree. However, it would still encrypt the (potentially not TLS encrypted) overlay traffic except for the headers.

@PlagueCZ had the point that, during the debugging session (even in the production), you also need to see the info provided in the inner header so as to understand the full trace/path of a packet.

ieps/NNNN-underlay-encryption-dpservice.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Alternatives
+
+- If no encryption is required, the additional graph nodes could simply be left out of the graph. However, since the graph creation relies on preprocessor commands, the configuration is fixed at compile time. It is possible to simply skip the operation of the nodes with application runtime configuration but the nodes would still be part of the graph. The "feature arc" functionality of the graph library seems to bridge that gap but it is still handled as experimental in the [documentation](https://doc.dpdk.org/api/rte__graph__feature__arc_8h.html) and apparently it comes with a performance impact even if the feature is not used ([DPDK summit presentation](https://youtu.be/AENsTjL_6eA?list=PLo97Rhbj4ceI3ENbGEN44mBVkLtdYB0DC&t=1493)).
+
+- If one wanted to change the scope of the encryption, in order to be able to see the encapped IPv4 headers, for example, it would in principle be possible to move the encrypt node in the graph in front of the encapsulate node. This way we'd end up with an encrypted IPv4 package with visible headers, that would then be encapped. I.e.:
```
byteocean, line 204: Is there a particular reason to move the encrypt node before the encapsulation node? it is possible to achieve that by shifting the starting pointer to the intended packet field and encrypt the content after this pointer?

ieps/NNNN-underlay-encryption-dpservice.md
```diff
@@ -0,0 +1,195 @@
...
+When replay protection is required, the sequence numbers of each packet must be known on both sides. The sender needs to increment it one by one and the receiver needs to keep track of the last numbers to keep his replay window up to date. On the receiver side, this state cannot be rebuilt purely from the information contained in one or multiple packets. Making a stateful tracking of packets necessary for both communication ends.
+
+In case statelessness is desired instead, [RFC4304 - 3.3.3](https://www.rfc-editor.org/info/rfc4303/#section-3.3.3) contains information on how ESP should be configured without replay protection. This configuration is on per SA basis, meaning that some SAs could be replay protected while others are not. DPDK's IPsec library allows enabling and disabling of sequence number checks via the [replay_win_sz](https://doc.dpdk.org/api/structrte__security__ipsec__xform.html#a85eb02412c31f7fbf81890c32a231b95). This makes it configurable whether this state should be built and maintained.
+
+#### Impact on HA abilites
```
byteocean, line 141: Although dpservice already has the HA capability but the upstream metalnet is missing the support for it. It seems reasonable to me not to implement this part in the first iteration unless further enhancement for this feature would drastically change the current design, e.g., data structure to be used.

ieps/NNNN-underlay-encryption-dpservice.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Alternatives
+
+- If no encryption is required, the additional graph nodes could simply be left out of the graph. However, since the graph creation relies on preprocessor commands, the configuration is fixed at compile time. It is possible to simply skip the operation of the nodes with application runtime configuration but the nodes would still be part of the graph. The "feature arc" functionality of the graph library seems to bridge that gap but it is still handled as experimental in the [documentation](https://doc.dpdk.org/api/rte__graph__feature__arc_8h.html) and apparently it comes with a performance impact even if the feature is not used ([DPDK summit presentation](https://youtu.be/AENsTjL_6eA?list=PLo97Rhbj4ceI3ENbGEN44mBVkLtdYB0DC&t=1493)).
```
byteocean, line 202: what we usually do for this case is to introduce a compilation flag that can enable/disable a feature. one of the examples is for 'virtual service' [option](https://github.com/ironcore-dev/dpservice/blob/58eb2049bbe47f6eeb15cb1c18b3489c9b85403a/meson_options.txt#L5).

ieps/NNNN-underlay-encryption-dpservice.md
mast-wch: I think the reason would be ease of use and sticking directly to the dpdk library functions.

You are right, one could manually create the crypto operation structure to point to the data of the encapsulated package. Currently this is completely managed by the dpdk ipsec library functionality. `rte_ipsec_pkt_crypto_prepare` finds the data, aad, and ICV locations in the buffer as long as the mbufs length fields and the `rte_crypto_sym_xform` struct are properly setup. If we were to use the same functionality on the unencapsulated package we'd achieve the same ease of use without manual crypto operation handling - we would just need to apply it before encapsulation takes place. 

What could work as well (and I'm not certain whether that is what you actually meant) could be to create an mbuf object for the encapsulated package which could be handled by the ipsec library as is and would yield an encrypted inner package. Then we would only have to manually adjust the outer ipv6 header for the changed package size of the inner package. Also a bit more effort but would also work after the encapsulation node.

ieps/NNNN-underlay-encryption-dpservice.md
mast-wch: I've added a reference and hope the motivation is now a bit less thin.

ieps/NNNN-underlay-encryption-dpservice.md
mast-wch: I'm feeling free enough to resolve this discussion as I can't add anything else and there seems to be no further questions in this regard.

ieps/NNNN-underlay-encryption-dpservice.md
mast-wch: As the problematic data structures seem to be those for the replay window and sequence numbers that are completely maintained by dpdk at this point, I don't think we'd change the data structures if it is not needed. Setting the replay window to 0 seems the defined way for getting rid of the replay protection and that sort of state. That is described in the text you marked above.  

Implementing the part that follows below (the sequence number batching) is more complicated, I assume. At least I have not yet added that to a working prototype and it is only a concept so far.

If your suggestion is to get started with the general implementation and implement HA capabilities in a later step, then I'd be fine with that and would reflect that in the document.

ieps/NNNN-underlay-encryption-dpservice.md
mast-wch: Looking at the PoC Guvenc introduced, the command line parameter plus flexible edge assignment seems to be a working option that avoids having multiple binaries (which would be the result of the command line flag option, I think). I'd stick to that one if it is not preferred to have compilation flags instead.

ieps/NNNN-underlay-encryption-dpservice.md
mast-wch: reflected both options in the text, closing the discussion as no further input was given

ieps/NNNN-underlay-encryption-dpservice.md
mast-wch: mentioned the possibility and problem in the iep, closing the discussion as no further input as given

ieps/NNNN-underlay-encryption-dpservice.md
mast-wch: mentioned the decision in the iep, closing the discussion as no further input was given

byteocean (approved):

maxmoehl (approved):

PR #51, opened 2026-07-22
Add proposal for key exchange for the underlay encryption

kitsudaiki:
This PR is one of the two replacement-proposals for the PR  https://github.com/ironcore-dev/enhancements/pull/38 
in order to provide a key exchange in metalbond for the underlay encryption of PR https://github.com/ironcore-dev/enhancements/pull/50

### DIFF
--- a/ieps/NNNN-underlay-encryption-key-exchange.md
+++ b/ieps/NNNN-underlay-encryption-key-exchange.md
@@ -1,0 +1,139 @@
+---
+title: Underlay Encryption - Key Exchange
+
+iep-number: NNNN
+
+creation-date: 2026-07-21
+
+status: implementable
+
+authors:
+
+- "@kitsudaiki"
+- "@mast-wch"
+- "@markus-hentsch"
+
+reviewers:
+
+- "tbd"
+
+---
+
+# IEP-NNNN: Underlay Encryption - Key Exchange
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Alternatives](#alternatives)
+
+## Summary
+
+Implementation of a key exchange between hosts to provide keys for the underlay encryption of PR https://github.com/ironcore-dev/enhancements/pull/50.
+
+## Motivation
+
+To provide keys for the underlay encryption in dpservice, symmetric keys have to be shared between hosts with dpservice.
+
+### Goals
+
+- Add a key exchange into Metalbond
+- Using the MLS protocol for the first variant for the key exchange
+
+### Non-Goals
+
+## Proposal
+
+### Related previous issues and PRs
+
+There are already 2 open issues related to this topic:
+
+- https://github.com/ironcore-dev/metalnet/issues/320
+- https://github.com/ironcore-dev/roadmap/issues/69
+
+There was also a first version for an enhancement document for the underlay encryption:
+
+- https://github.com/ironcore-dev/enhancements/pull/38
+
+This was closed, because as a result of the community discussion, the originally intended use of IKEv2 was replaced by MLS. Additionally, it was decided to split it into 2 smaller enhancement proposals.
+
+### Overview
+
+```mermaid
+graph TD
+    metalnet01["Metalnet"]
+    metalnet02["Metalnet"]
+
+    dpservice01["dpservice"]
+    dpservice02["dpservice"]
+
+    mls_client_plugin01["MLS-Client-Plugin"]
+    mls_client_plugin02["MLS-Client-Plugin"]
+
+    subgraph Controller
+        subgraph metalbond_server["Metalbond-Server"]
+            mls_server["MLS-Server-Plugin"]
+        end
+    end
+
+    subgraph Compute Host 02
+        subgraph metalnet01["Metalnet"]
+            subgraph metalbond_client01["Metalbond-Client"]
+                mls_client_plugin01 <--> |TCP|mls_server
+            end
+            metalbond_server <--> |TCP|metalbond_client01
+        end
+        metalnet01 --> |gRPC| dpservice01
+    end
+
+    subgraph Compute Host 01
+        subgraph metalnet02["Metalnet"]
+            subgraph metalbond_client02["Metalbond-Client"]
+                mls_client_plugin02 <--> |TCP|mls_server
+            end
+            metalbond_server <--> |TCP|metalbond_client02
+        end
+        metalnet02 --> |gRPC| dpservice02
+    end
+
+    %% dpservice01 <--> |IPSec| dpservice02
+```
+
+### Selected protocol
+
+In the [previous PR](https://github.com/ironcore-dev/enhancements/pull/38) there was a discussion about the method, 
+how to share the key for the encryption. 
+In the end the conclusion was to use MLS (<https://datatracker.ietf.org/doc/rfc9420/>) as first implementation, instead of IKEv2.
+
+The [OpenMLS](https://github.com/openmls/openmls) library is so far the best maintained implementation of the MLS protocol, so it is used for this first version of the underlay encryption. 
+Because OpenMLS is written in Rust, a small glue layer has to be implemented to expose the necessary functions to the Go code. 
+
+### Metalbond
+
+To avoid a new CRD for the SA (Security Association) as well as a resulting CR for each connection and because it also has to work on hosts without Metalnet, 
+the key exchange will be integrated into Metalbond. 
+The MLS protocol differs quite a lot from the Metalbond protocol. 
+It would be a very big change to modify the Metalbond protocol to also make all MLS traffic possible over it. 
+Besides this, this would mean MLS-specific extensions into the protocol. 
+This would result into a hard commitment for the MLS protocol for the future. 
+To avoid this problem of MLS-specific modifications, the MLS traffic will be handled by a second parallel TCP connection with a MLS server next to the Metalbond server. 
+Additionally all MLS-related features on server and client side will be encapsulated into a MLS plugin, 
+which will be used by Metalbond. There will be 2 plugins, one for the client-side and one for the server-side. 
+This way it will be easily possible in the future to exchange MLS for example by an IKEv2 plugin, if necessary. 
+So there is no hard commitment for the usage of MLS to share keys over different hosts.
+
+### Metalnet
+
+Metalnet also has to be extended:
+
+1. The key, salt and SPI, coming from the key exchange from Metalbond, 
+   have to be forwarded to dpservice. The key will be pushed into dpservice, before the route is created by Metalbond.
+
+2. A new bool property has to be added to the Network CR to tell Metalnet whether the network has to be encrypted or not.
+
+## Alternatives
+
+- Use a Go-native MLS implementation like https://github.com/thomas-vilte/mls-go or https://github.com/trevex/mls-go instead of the OpenMLS Rust implemenation with the glue layer. However, these repositories seem to be much less actively maintained than OpenMLS.


ieps/NNNN-underlay-encryption-key-exchange.md
```diff
@@ -0,0 +1,139 @@
...
+
+Implementation of a key exchange between hosts to provide keys for the underlay encryption of PR https://github.com/ironcore-dev/enhancements/pull/50.
+
+## Motivation
+
+To provide keys for the underlay encryption in dpservice, symmetric keys have to be shared between hosts with dpservice.
+
+### Goals
+
+- Add a key exchange into Metalbond
```
trevex: Personally I believe the key exchange should be initially developed independently (not outside of ironcore-dev, but not as part of metalbond).

With minor additions to metalnet, this can be seamlessly realized as addon:
1. An `encrypted` field marks a NIC as encrypted.
2. If a NIC is `encrypted` the control plane waits for an "encryption ready" condition.

With these available an independent set of components could drive the key management and just install SAs as necessary.

MLS requires strong consistency guarantees. If it is redundantly operated like metalbond, we double the amount of SAs by effectively joining count of metalbond groups per VNI.

We should iterate independently first and find the best solution for key exchange and decide after whether metalbond and the key exchange should be homogenized.

kitsudaiki: with a `encryption ready` status in the NetworkInterface-CR you have no guarantee, that every participant has to key already pushed into its dpservice. Who should set it? If only the one, who announced the route sets it and knows by the status-field, that the key is ready, the other participants can not be sure that they have already received the key. You would have to track every peer of a route by its own status-field and the idea of a new CR per connection-pair to do this was discarded in the community meeting, because of scaling concerns. The idea to put it into metalbond was to make sure that it is in sync and everyone can be sure, when the key is available to start the communication. It was also a requirement in the meeting, that the MLS works on hosts with Metalbond-client but with no Metalnet, which also leaded to the motivation to put this into Metalbond itself.

But yes, the point with the redundantly operated metalbond is a valid point. But placing the server component of MLS into its own container is the least problem.

trevex: > with a encryption ready status in the NetworkInterface-CR you have no guarantee, that every participant has to key already pushed into its dpservice. 

This is not how MLS works from my understanding. MLS is a protocol using a blind middleman for key exchange. metalbond would therefore be blind to the keys anyway.
Each metalnet would join for their local virtual NICs the relevant MLS groups. The key exchange is stateful, so each peer knows when a key for the group is committed, when it is rotated etc.

> You would have to track every peer of a route by its own status-field and the idea of a new CR per connection-pair to do this was discarded in the community meeting, because of scaling concerns.

I am not sure I quite understand, if an agent installs the key after the MLS group was successfully joined and committed, then this is also true for other participants. From my understanding the desired model is MLS group SA, which is used to derive distinct SPIs and salt to avoid seq no collisions, so no other CRs are required.

By introducing minor changes to the core API we can allow key exchange addons to drive encryption completely independent.

trevex: If we compare this with IKE, we also know when the key exchange is established (and key is installed in dpservice) and can signal an encryption ready condition.

kitsudaiki: updated the document. hope our concept is now more clear and makes more sense to you. Please let us know, if you still see some major problems.

ieps/NNNN-underlay-encryption-key-exchange.md
```diff
@@ -0,0 +1,193 @@
...
+
+Implementation of a key exchange between hosts to provide keys for the underlay encryption of PR https://github.com/ironcore-dev/enhancements/pull/50.
+
+## Motivation
+
+To provide keys for the underlay encryption in dpservice, symmetric keys and all other necessary information have to be shared between hosts with dpservice.
+
+### Goals
+
+- Add a key exchange into Metalbond without deep integration into Metalnet, to also support hosts without Metannet
```
byteocean: that is not true anymore or? because the key exchanging work is offloaded to the implemented  MLS server as described below.

ieps/NNNN-underlay-encryption-key-exchange.md
```diff
@@ -0,0 +1,193 @@
...
+- Using the MLS protocol for the first variant for the key exchange, with OpenMLS as specific implementation of this protocol
+- One MLS-group/Key per route, to be in sync with the Metalbond workflow and to support encrypted connections between different VNIs
+- Only support for dpservice in the current state of the proposal
```
byteocean: dpservice is out of the scope of this proposal, IMO

ieps/NNNN-underlay-encryption-key-exchange.md
```diff
@@ -0,0 +1,193 @@
...
+### Selected protocol
+
+In the [previous PR](https://github.com/ironcore-dev/enhancements/pull/38) there was a discussion about the method, 
+how to share the key for the encryption. 
+In the end the conclusion was to use MLS (<https://datatracker.ietf.org/doc/rfc9420/>) as first implementation, instead of IKEv2.
+
+The [OpenMLS](https://github.com/openmls/openmls) library is so far the best maintained implementation of the MLS protocol, so it is used for this first version of the underlay encryption. 
+Because OpenMLS is written in Rust, a small glue layer has to be implemented to expose the necessary functions to the Go code. 
+
+### Metalbond
```
byteocean: I would suggest to rename this section to "MLS server" and move the content of basic flow under it.

ieps/NNNN-underlay-encryption-key-exchange.md
kitsudaiki: yes, it is a bit misleading. just wanted to describe, that there are only code-changes in metalbond to interact with the key-exchange. changed the text to `Couple key exchange only to Metalbond`.

ieps/NNNN-underlay-encryption-key-exchange.md
kitsudaiki: yes. It is. It was meant, that NetlinkClient is not supported in context of this proposal. changed it to reference the other PR with the dpservice encryption and its gRPC-extension, where this implementation will be attached to.

ieps/NNNN-underlay-encryption-key-exchange.md
kitsudaiki: good point, but moved it to the overview. I think this would be a better fit.

kitsudaiki:
changed it back to Draft for the moment to make some changes:
- make bundling part of the current proposal instead of an idea for future enhancements
- add key-rotation workflow
- decouple metalbond-client and the key-exchange by another grpc-channel to make the key-exchange easier replaceable by other solutions in the future

kitsudaiki:
Updated proposal:
- use the IP-Prefix instead of the single IP's
- decoupled metalbond-code and the key-exchange by another gRPC-channel to make the key-exchange easier replaceable by other solutions in the future
- were able to remove the back-channel
- now MLS-group per VNI instead of Route

ieps/NNNN-underlay-encryption-key-exchange.md
```diff
@@ -0,0 +1,233 @@
...
+
+        metalbond_client02 --> |gRPC| mls_client_02
+        metalbond_client02 --> |gRPC| dpservice_02
+        mls_client_02 --> |gRPC| dpservice_02
+    end
+
+```
+
+To avoid a new CRD for the SA (Security Association) as well as a resulting CR for each connection and because it also has to work on hosts without Metalnet, 
+the key exchange will be coupled with Metalbond.
```
byteocean, line 123: the statement should be the opposite as there exists an independent channel for MLS message exchanging?

markus-hentsch: @byteocean I think it is just phrased in a slightly misleading fashion. What do you think of this rephrasing:

```suggestion
the key exchange will be initiated by the existing Metalbond workflows but happen on its own dedicated channel.
```

byteocean: yes, I also thought it was just a phrasing issue. and the above change sounds clear enough. thanks.

kitsudaiki: applied the suggestion of markus

ieps/NNNN-underlay-encryption-key-exchange.md
kitsudaiki: because of the major changes in the meantime, this thread is outdated and will be closed

ieps/NNNN-underlay-encryption-key-exchange.md
kitsudaiki: because of the major changes in the meantime, this thread is outdated and will be closed

ieps/NNNN-underlay-encryption-key-exchange.md
kitsudaiki: because of the major changes in the meantime, this thread is outdated and will be closed

ieps/NNNN-underlay-encryption-key-exchange.md
kitsudaiki: because of the major changes in the meantime, this thread is outdated and will be closed

kitsudaiki:
set iep number 25 and added @byteocean  to the rewiever field. hope this is OK and complete now

ieps/25-underlay-encryption-key-exchange.md
```diff
@@ -0,0 +1,233 @@
...
+To avoid a new CRD for the SA (Security Association) as well as a resulting CR for each connection and because it also has to work on hosts without Metalnet, 
+the key exchange will be initiated by the existing Metalbond workflows but happen on its own dedicated channel.
+The MLS protocol differs quite a lot from the Metalbond protocol. 
+It would be a very big change to modify the Metalbond protocol with MLS-specific extensions to make all MLS traffic possible over it. 
+This would result into a hard commitment for the MLS protocol for the future. 
+To avoid this problem of MLS-specific modifications, the MLS traffic will be handled by a second parallel HTTP connection with a MLS server next to the Metalbond server. 
+
+OpenMLS is written in Rust. Both communicate with a gRPC connection. The Go code handles the orchestration 
+(subscriptions, key rotation, ...) and the Rust code is just a wrapper for the OpenMLS and handles only the MLS groups and members, 
+bases on the messages coming from the Go code.
```
byteocean: bases on -> based on

kitsudaiki: have fixed it

ieps/25-underlay-encryption-key-exchange.md
```diff
@@ -0,0 +1,233 @@
...
+The MLS protocol differs quite a lot from the Metalbond protocol. 
+It would be a very big change to modify the Metalbond protocol with MLS-specific extensions to make all MLS traffic possible over it. 
+This would result into a hard commitment for the MLS protocol for the future. 
+To avoid this problem of MLS-specific modifications, the MLS traffic will be handled by a second parallel HTTP connection with a MLS server next to the Metalbond server. 
+
+OpenMLS is written in Rust. Both communicate with a gRPC connection. The Go code handles the orchestration 
+(subscriptions, key rotation, ...) and the Rust code is just a wrapper for the OpenMLS and handles only the MLS groups and members, 
+bases on the messages coming from the Go code. 
+
+MLS groups are created per VNI. The first one, who subscribes to a VNI, also creates the group for it. Everyone, who subscribes afterwards to the VNI, is automatically invited into the group and receives the group key.
```
SamiraBriongos: This decision makes it possible for any host in one VNI to decrypt the messages for any SA on the same VNI. So it assumes all the members are then trusted and allowed to do so. It would be great if you could add a couple of lines explaining the threat model so it is clear to everyone.

kitsudaiki: It was basically a design decision by the community. Our first approach was to make 1 MLS-group and so one Key per Route instead of per VNI. The community decided that these are too many MLS-groups in the end and they want only one key per VNI.

kitsudaiki: I agree with you, that 1 key per VNI has security concerns, but that was not the decision of me. The community had concerns regarding the scalability, so 1 key per VNI instead of per Route was a compromise.

ieps/25-underlay-encryption-key-exchange.md
```diff
@@ -0,0 +1,233 @@
...
+Because the addresses are pairwise numerically sorted and because each host knows his own prefix address, 
+every connection pair of two hosts is capable of independently deriving the identical SPI and salt for their specific connection.
+To make this work, each host must have a unique address prefix in the underlay network.
+
+All key exchange related parts are separated from the Metalbond and Metalnet code by a minimal gRPC connection. This makes it exchangeable more easily.
+
+The periodical key rotation is only done internally without any intervention from the outside by Metalbond or Metalnet. 
+The MLS server triggers the key rotation and ensures, that all clients on the hosts of the group are in sync, 
+to avoid deleting an outdated key before every client received the new one. 
+A key rotation always results in a new salt and SPI too.
```
SamiraBriongos: The MLS server deserves more attention. Considering that the salts are deterministically derived, and that  nonce reuse leaks both plaintext and key material, it should be guaranteed that after a crash or a restart of the server, it never recovers from a snapshot older than the latest committed epoch. Also, consider including some fork protection to prevent that either a second redundant MLS server is started or if you need it, this second server should not be able to commit an epoch already used by the primary one. 
Ideally, include a short section on the crash-and-recovery strategy of the planned MLS server.

ieps/25-underlay-encryption-key-exchange.md
SamiraBriongos: The decision is ok, the scalability tradeoff makes sense. However, since this choice determines who can decrypt what, could you include a short sub-section on the threat model: which adversaries the proposal protects against, and where the trust boundaries lie? Stating this explicitly would make the compliance claims checkable against a defined trust domain and would make it obvious to readers which risks are accepted by design.

ieps/25-underlay-encryption-key-exchange.md
kitsudaiki: added a new paragraph at the end of the section regarding your concerned reuse-case in the MLS-server

ieps/25-underlay-encryption-key-exchange.md
kitsudaiki: we have added a Threat model section to the Motivation of the document.

byteocean (approved):

PR #55, opened 2026-07-27
Add Server Parked State enhancement proposal

### LINKED ISSUE #44: Typed announce/approve contract with deferral
## Summary

Replace the stringly-typed maintenance label/annotation flow (`metal.ironcore.dev/maintenance-{needed,requested,approved,reason}`) with a typed announce/approve contract on `Server` and `ServerClaim`. The consumer (the only party that knows which servers form a cohort) coordinates disruption *timing* by approving announces serially and **deferring** the rest; metal-operatorowns the contract plus a blunt concurrency cap, not a fleet-quorum budget.

## Motivation

The current handshake is carried entirely through labels/annotations on `Server` plus `ServerMaintenanceSpec.Policy` (`OwnerApproval` vs. `Enforced`). This works for the happy path but:

- The handshake is invisible to the type system: no validation, no status, no way for an external actor (CAPI/MCM, a workload manager) to observe where a request is in the announce -> approve -> drain lifecycle without scraping strings.
- Approval and policy are conflated with the maintenance object, so the negotiation between the `Server`/claim owner and the maintenance actor has no first-class representation.
- There is no fleet-wide rate limiting: `Enforced` can evict unboundedly in a burst; `OwnerApproval` can block forever on one unresponsive owner.

## Goals

- Typed announce/approve fields/conditions on `Server` and `ServerClaim` replacing the `metal.ironcore.dev/maintenance-*` labels and annotations.
- A **deferral** response in the handshake ("not now, re-ask after T"), so the consumer can serialize disruption within a cohort without metal-operator knowing the cohort exists.
- A global **concurrency cap** on in-flight evictions as blast-radius / runaway protection: a cap, not a semantic budget.
- Emergency override: long-lived bound claims with no responding consumer escalate to enforced after a grace period, subject to the concurrency cap.
- Documented deprecation path for `ServerMaintenanceSpec.Policy` (`OwnerApproval` / `Enforced`).

## Non-Goals

- Redesigning the `Server` state machine.
- Moving `ServerMaintenance` out of `metal-operator` (WS 3): this workstream defines the contract, not where its controllers live.
- Cordon (WS 1): cordon prevents new bindings; this governs drain/eviction of existing bindings.
- The CAPI/MCM integration itself (WS 6): this defines the contract that integration consumes.

afritzler:
Add Server Detachment State enhancement proposal.

Fixes #44


adracus (approved):

ieps/23-server-parked-state.md
```diff
@@ -0,0 +1,153 @@
...
+   resume.
+
+The critical requirement: **during step 2, neither reconciler may boot the
+claim image** on an intermediate restart nor "heal" the power state back to
+what the claim wants; both controllers stand down until an explicit "bring
+back" signal is received.
+
+### Goals
+
+- Add a `Parked` overlay state to the `Server` state machine that suspends
```
atd9876, line 91: How do we handle a server with the OS installed to local disk and the Bios Boot Order updated by the OS install to Hdd first. When it is powered on by the external actor it will boot to that production OS without any interaction with the metal-operator.

I would suggest that a continuous boot override to network, but then the boot-operator would have to be Park aware and return a 404 type error or return the info to boot from some prescribed image for this situation.

afritzler: @adracus is currently working on https://github.com/ironcore-dev/metal-operator/issues/1052. This should be the component you can use to do the boot going forward without `metal-operator` and `boot-operator` in the loop.

Regarding the intermediate boot: I would suggest that the external party will have to ensure a correct boot order for whatever activity it want to do.

afritzler: The idea behind the `Parked` state is, that the `metal-operator` gives away control over the `Server` and `ServerClaim` resources.

xkonni (approved):
I can see some edge cases where this might be problematic, like what @atd9876 commented, but like the overall concept and think we need something like this.

Idea: Optional wipe before parking?

afritzler:
Well the idea of parking is first of all to keep the `ServerClaim` around - wiping the Server will be basically: delete the claim + set the reclaim policy to Retain and run the sanitization.

afritzler:
Introduced a two-annotation model for parking: a transient `metal.ironcore.dev/operation: park` request that is removed once the server reaches `Parked`, and a durable internal `metal.ironcore.dev/parked` annotation that holds the parked state so it can always be reconstructed and the reconcilers keep standing down across restarts. Resume is driven by the requestor removing the parked annotation, transitioning back to `Reserved`/`Available` based on the prior state.

adracus (approved):

maxmoehl:
> Introduced a two-annotation model for parking: a transient `metal.ironcore.dev/operation: park` request that is removed once the server reaches `Parked`, and a durable internal `metal.ironcore.dev/parked` annotation that holds the parked state so it can always be reconstructed and the reconcilers keep standing down across restarts. Resume is driven by the requestor removing the parked annotation, transitioning back to `Reserved`/`Available` based on the prior state.

Why not just make the `operation: park` annotation durable? We already do that for `ignore` and I'm not sure I like the two annotation dance.

adracus:
The `ignore` annotation is more of an anti-pattern: It's not an operation in the sense of 'please do something now' but rather an instruction towards the reconciler to not do any reconciliation work on a server whatsoever.

In contrast, parking is a real operation: It should gracefully shut down any running server claims and only once that has happened update its status to parked. For the reconciler to be able to know wheter it successfully parked itself, the 'internal' state annotation is required, as per k8s api conventions we should not store internal state information on a resources `.status.state`.

Then, when a user wants to 'unpark' a server, the same thing happens again: The user requests unparking via the `unpark` operation annotation, the reconciler drives that and only once done sets the `status.state` and clears the internal state annotation.

afritzler:
The only point worth discussing here is: do we need two annotations?

1. If we want to use `operation: park` annotation -> then I would say yes we need two. That way we can use the operation annotation as a transient "hey I want to park this server" signal.
2. If we only want to use 1 then the `parked` annotation can be used as a durable signal to the machinery. You can transition the server, reconstruct the state + the requester only has one annotation to deal with. Downside: every reconciliation needs to evaluate what `parked` means like: did I already shutdown the server or not.

maxmoehl:
> The only point worth discussing here is: do we need two annotations?

Agreed.

> 1. If we want to use `operation: park` annotation -> then I would say yes we need two. That way we can use the operation annotation as a transient "hey I want to park this server" signal.

I'd prefer 2.

> 2. If we only want to use 1 then the `parked` annotation can be used as a durable signal to the machinery. You can transition the server, reconstruct the state + the requester only has one annotation to deal with. Downside: every reconciliation needs to evaluate what `parked` means like: did I already shutdown the server or not.

Well, if `status.state != Parked` you have to check if its powered off and those things to then set the state to parked, but if `status.state == Parked` you can skip all of that as you know it was already done. Essentially the status becomes a cache of sorts which seems to be what it's for anyway, no?

adracus:
But again, this is an anti-pattern: Using `.status.state` for a reconciler of a resource to drive the reconciliation should not be done. `status` of a resource should be (from the perspective of the resource itself) wipable at any time.

maxmoehl:
We had a discussion in the office on this topic and back-tracked a bit. I'm still opposed to the notion of having a server "suspended", e.g. a claim is still attached but the server is used otherwise for maintenance and then returned to the claim at some point in the future.

This could go wrong in so many ways and being explicit (by detaching and re-attaching a claim) is my preferred option. If you have state on the server, e.g. some ceph disks, you should never assume that the data is "just there", for example a disk swap would invalidate that. You are of course free to store any state and _try_ to gracefully recover it, but this is an explicit thing, not something that "just happens" to a claim. Also the taints and tolerations system has a clear path to gracefully and forcefully control the flow of operations whereas parked is just forced upon whoever owns the claim.

afritzler:
What this proposal explicitly doesn't cover is the "approval flow". So the force upon whoever owns the claim argument doesn't really apply here. Of course you need to negotiate with the claim owner that you need to do maintenance on his Server.

Also, of course you could do something like that: 
1. cordon/taint the server 
2. power down the server 
3. add the `ignore` annotation to both server and claim

IMO this is a very brittle "contract" + also has no clear state that the server is now under maintenance by somebody else. Also you need to reverse the flow if you want to "un-park" the server.

Of course there is a lot of stuff which can go wrong during those maintenance operations. If the maintenance operation will be destructive so workload eviction has to be done upfront and maybe even the claim has to be deleted. But in that case I would then cordon the server to forbid further claiming. Even in that case where the server is available and nobody is using it I want to have a clear path to do day-2 (of course I could `ignore` the server).

But imagine a non-evictable workload like a big in-memory database. You don't want to evict that from a server just because you want to update a BMC firmware.

ieps/23-server-parked-state.md
```diff
@@ -0,0 +1,195 @@
...
+   This is a one-shot request; it does **not** itself persist the parked state.
+2. **Park.** The `Server` reconciler observes the request and:
+  - powers the server **off** (idempotent; only if not already off),
+  - ensures the claim's `BootConfigurationRef` is **not** on the active boot
+     path so an intermediate restart cannot boot the claim image,
+  - records the parked state by setting the **internal** annotation
+     `metal.ironcore.dev/parked: "true"`,
+  - sets `status.state = Parked`,
+  - **removes** the `metal.ironcore.dev/operation: park` request annotation
+     again, so the request does not linger once the server has reached `Parked`.
```
nagadeesh-nagaraja, line 136: do we allow further power operation on the server?

we still need a way to control the power of server during this external maintenance. 

do we allow reboot operation on the server in parked state??

afritzler: The 3rd party requesting to park the server takes over the power management. The `metal-opearator` stands down on all operations against this `Server`/`ServerClaim`.

stefanhipfel:
@afritzler thx for creating that proposal. 
How about extending the serverMaintenance with a strategy?

```
apiVersion: metal.ironcore.dev/v1alpha1
kind: ServerMaintenance
metadata:
  name: maintenance-server-01
spec:
  serverRef:
    name: server-01
  strategy: Park # Options: SoftMaintenance, Park, FirmwareUpdate...or whatever names we come up with
  powerState: Off
```
In case of `strategy: Park` both claim and server controllers stand down.

afritzler:
@stefanhipfel since we will move the `ServerMaintenance` resource to the `metal-maintenance-operator` we will re-define the maintenance contract. Therefore we can do something like that there. I would keep the current `ServerMaintenance` type as is in the `metal-operator` as this is now deprecated. 

The idea behind `cordon`/`Taints` and now this `Parked` state is, that you have from the outside a clear contract to interface with the `Server` resource. The `metal-operator` stays clean of any day-2 tasks but offers enough API surface for external parties to model any day-2 flow.

